PR #10, opened 2022-12-24
Document Specification Draft v0.0.0

jlbabilino:
Includes the initial draft of the specification for __Documents__.


JoshuaAN (approved):

shueja-personal (commented):
Can we do empty arrays for the no-points trajectory and waypoints lists?

PR #18, opened 2022-12-26
switch to css modules to enable browser

shueja-personal:
Apparently the importable CSS modules need a declared TS type or everything screams.


shueja-personal:
Merge after #12

jlbabilino (approved):

PR #86, opened 2023-07-25
Add limit of 3x zoom on field

shueja-personal:


jlbabilino:
I think this limit is too strict. I should be able to zoom out so the field is about three times smaller than the width of the viewport.

src/components/field/svg/FieldOverlayRoot.tsx
```diff
@@ -43,6 +43,7 @@ class FieldOverlayRoot extends Component<Props, State> {
     this.handleResize();
     let zoomBehavior = d3
       .zoom<SVGGElement, undefined>()
+      .scaleExtent([1, 3])
```
jlbabilino, line 46:
```suggestion
      .scaleExtent([0.3, 12])
```

jlbabilino: This is what I tried and like. Lmk what you think.

jlbabilino (approved):

PR #93, opened 2023-07-29
UI Overhaul, Document Spec Migration

shueja-personal:
DRAFT until otherwise noted.

PR #94, opened 2023-07-31
rounding

Autumn-Ou:
rounding in waypoint editing and paused time slider bar


shueja-personal:
Closed since both issues are solved in #93

PR #108, opened 2023-08-06
Use platform-specific tauri config

jlbabilino:
Before this, I was manually modifying the `tauri.conf.json` file based on the platform.

This PR uses [this feature](https://tauri.app/v1/api/config/#platform-specific-configuration) instead of all that nonsense.


shueja:
Is this still worth doing?

PR #114, opened 2023-09-17
Add json schema for documents

Lewis-Seiden:


src/document/DocumentManager.ts
```diff
@@ -116,6 +119,10 @@ export class DocumentManager {
 
   async saveFile() {
     const content = JSON.stringify(this.model.asSavedDocument(), undefined, 4);
+    if (!VERSIONS["v0.0.1"].validate(this.model.asSavedDocument())) {
+      console.error("Invalid Doc JSON: " + "\n" + JSON.stringify(documentSchema) + "\n" + content);
+      return;
+    }
```
jlbabilino, line 125: Is the purpose of this for developers to check that their documents are saved with the right schema? For example, if they messed something up in the document model, it would no longer be spec compliant, and this would let them know? If that's the case, I'd make this a warning instead of an error.

jlbabilino: It's most important for this to be in the open function in addition to save.

src-tauri/Cargo.toml
```diff
@@ -15,7 +15,7 @@ license = "BSD-3-Clause"
 tauri-build = { version = "1.4", features = [] }
 
 [dependencies]
-tauri = { version = "1.4", features = [ "dialog-save", "dialog-open", "fs-all", "shell-open"] }
+tauri = { version = "1.4", features = [ "dialog-save", "dialog-open", "fs-all", "shell-open", "devtools"] }
```
jlbabilino, line 18: Make sure to remove this before merging.

src/document/DocumentManager.ts
```diff
@@ -1,8 +1,9 @@
 import { createContext } from "react";
 import StateStore, { IStateStore } from "./DocumentModel";
 import { dialog, fs } from "@tauri-apps/api";
-import { v4 as uuidv4 } from "uuid";
-import { applySnapshot, getRoot, onPatch } from "mobx-state-tree";
+import { v4 as uuidv4, validate } from "uuid";
```
shueja-personal, line 4: You probably didn't mean to import validate here.

src/document/DocumentManager.ts
```diff
@@ -67,7 +68,13 @@ export class DocumentManager {
   }
   async onFileUpload(file: File | null) {
     await this.parseFile(file)
-      .then((content) => this.model.fromSavedDocument(JSON.parse(content)))
+      .then((content) => {
+        if (VERSIONS["v0.1"].validate(JSON.parse(content))) {
```
shueja-personal, line 72: I forget exactly how the plumbing goes here but validation (using the correct document version) ought to happen before the automatic document upgrade chain

shueja-personal (changes requested):

src/document/DocumentSpecTypes.ts
```diff
@@ -92,11 +100,19 @@ export let VERSIONS = {
...
+      const ajv = new Ajv();
+      return ajv.validate(v0_0_1_Schema, document);
+    },
   },
   "v0.1": {
     up: (document: any): v0_1 => {
       return document as v0_1;
     },
+    validate: (document: v0_1): boolean => {
+      const ajv = new Ajv();
```
shueja-personal, line 113: Not sure if we need to make a new instance on every call but it's fine I think.

shueja-personal (approved):

jlbabilino (approved):

PR #148, opened 2023-11-15
Feature/control interval count heuristic

Lewis-Seiden:
depends on https://github.com/SleipnirGroup/TrajoptLib/pull/63


src-tauri/Cargo.toml
shueja, line 1: Of course, change this back after TrajoptLib#63

shueja (approved):

PR #162, opened 2023-11-17
added rotation guess heuristic

Lewis-Seiden:
not totally sold on the way i set this up yet, and even still could probably tune the heading weight a bit


shueja (commented):
You're missing a few absolute values here. What happens if dtheta is negative?

shueja (approved):

PR #177, opened 2023-12-02
save/export as .chor/.traj files

shueja:


Lewis-Seiden (commented):
do we want support for loading old files saved as .json?

shueja:
This still does, since the "open file" input doesn't do pattern checking.

Lewis-Seiden (approved):

PR #200, opened 2023-12-13
Port ChoreoLib to Python

gerth2:
Draft for now. Still need to add CI publishing to pypi


choreolib/py/choreolib/choreoTrajectory.py
```diff
@@ -0,0 +1,137 @@
...
+        heading: float,
+        velocityX: float,
+        velocityY: float,
+        angularVelocity: float,
+    ):
+        self.timestamp = timestamp
+        self.x = x
+        self.y = y
+        self.heading = heading
+        self.velocityX = velocityX
```
calcmogul, line 27: These names should ideally be snake case to follow https://pep8.org/#naming-conventions, though robotpy doesn't follow that standard.

TheTripleV: robotpy mostly matches java for wrapped stuff /shrug

shueja (commented):
I don’t know if the build setup is right but the actual code looks good

gerth2:
Added support to publish to pypi proper, same as photonvision has. Note neither has been tested in the main server, only in a different test environment. If we're willing to slap down another tag before kickoff we can see if it publishes correctly? I can always delete pypi things on my end if it goes bad.

image

shueja (commented):
Seems to be missing path following commands, is this due to robotpy's pending rewrite of commands implementation?

gerth2:
> Seems to be missing path following commands, is this due to robotpy's pending rewrite of commands implementation?

Possibly - I'll be honest I haven't revisited this till today. I've not attempted to keep the python implementation in sync with other languages.

I'm ok with abandoning this if it's not a good idea to scope creep at the moment. For any python team, they can always just clone and own a few python files into their codebase. Having official python support won't be a blocker for us using Choreo.

choreolib/py/choreolib/choreoTrajectory.py
TheTripleV: ctre has chosen to do snake case for their python stuff so I guess it's whatever now

.github/workflows/python.yml
```diff
@@ -0,0 +1,61 @@
...
+
+      - name: Publish package distributions to PyPI
+        # Only upload on tags
+        if: startsWith(github.ref, 'refs/tags/v')
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          packages_dir: ./choreolib/py/dist/
+
+    permissions:
+      id-token: write # IMPORTANT: this permission is mandatory for trusted publishing
```
spacey-sooty, line 61: Doubled?

choreolib/py/buildAndTest.bat
```diff
@@ -0,0 +1,14 @@
+:: Uninstall if it already was installed
```
spacey-sooty, line 1: If this is necessary it should have an equivalent POSIX compliant script

gerth2:
All good stuff. I think this one should stay paused till later this summer. Ideally, this needs to go through the robotpy wrappering process - only thing to pause on is making sure the choreolib C++ implementation is stable-ish (no new wonky dependencies, primarily). 

On my list to look at at least. The main roadblock is my own time and skill issues. But both are resolvable.

.github/workflows/python.yml
```diff
@@ -0,0 +1,61 @@
...
+  buildAndDeploy:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          sparse-checkout-cone-mode: false
+          fetch-tags: true
+          fetch-depth: 99999
```
auscompgeek, line 24: wat

.github/workflows/python.yml
calcmogul: I didn't actually need this permissions line for Sleipnir's Python package publishing.

calcmogul:
The setup.py needs to be ported to pyproject.toml; the former is deprecated by setuptools.

calcmogul:
I think the main thing left to do is add docstrings.

choreolib/py/pyproject.toml
```diff
@@ -0,0 +1,32 @@
...
+requires = [
+  "robotpy-wpimath",
+  "setuptools>=61.0",
+  "setuptools-git-versioning"
+]
+build-backend = "setuptools.build_meta"
+
+[tool.setuptools-git-versioning]
+enabled = true
+version_callback = "choreolib.version:get_version"
```
auscompgeek, line 28: Is there a particular reason to not use setuptools_scm?

calcmogul: It broke when I tried to use it for wpiformat.

calcmogul:
Docstrings are added. This needs new .traj files though, because module x and y forces were added to the output. Also, file deserialization and interpolation need to be implemented for those two fields.

calcmogul (approved):

PR #202, opened 2023-12-20
Choreolib: support other controllers for ChoreoControlFunction

tandej:
I'm not entirely sure what these other controllers would be, but only accepting WPI PIDControllers struck me as unnecessarily specific.


calcmogul:
Needs a C++ port.

PR #205, opened 2023-12-24
Circular Obstacles

Lewis-Seiden:


src-tauri/src/main.rs
```diff
@@ -233,8 +258,17 @@ async fn generate_trajectory(path: Vec<ChoreoWaypoint>, config: ChoreoRobotConfi
           }
         ]
       };
-    //path_builder.set_bumpers(config.bumperLength, config.bumperWidth);
-    path_builder.sgmt_circle_obstacle(0, path.len()-1, 3.0, 3.0, 1.0);
+
+    path_builder.set_bumpers(config.bumperLength, config.bumperWidth);
+
+    for o in circleObstacles {
+      println!("{:?}", o);
```
shueja, line 265: To Remove (the println)

src/components/sidebar/SidebarObstacle.tsx
```diff
@@ -0,0 +1,75 @@
...
+    let selected = this.props.obstacle.selected;
+    return (
+      <div
+        className={styles.SidebarItem + (selected ? ` ${styles.Selected}` : "")}
+        onClick={() => {
+          this.context.model.uiState.setSelectedSidebarItem(obstacle);
+          this.context.model.uiState.setSelectedNavbarItem(9);
+        }}
+      >
+        {React.cloneElement(<Circle></Circle>, {
```
shueja, line 43: The icon here is a circle, but is a circular (/) sign in the navbar. Use the latter here. Replace `<Circle></Circle>` with `ObstacleData["CircleObstacle"]`.

src/components/sidebar/SidebarObstacle.tsx
```diff
@@ -0,0 +1,75 @@
...
+        <span
+          className={styles.SidebarLabel}
+          style={{ display: "grid", gridTemplateColumns: "1fr auto auto" }}
+        >
+          {"Obstacle x: " +
+            this.props.obstacle.x.toFixed(2) +
+            ", y: " +
+            this.props.obstacle.y.toFixed(2)}
+        </span>
+        <Tooltip disableInteractive title="Delete Waypoint">
```
shueja, line 58: Tooltip here needs changing.

src/components/sidebar/SidebarObstacle.tsx
```diff
@@ -0,0 +1,75 @@
...
+          className: styles.SidebarIcon,
+          htmlColor: this.state.selected
+            ? "var(--select-yellow)"
+            : "var(--accent-purple)",
+        })}
+        <span
+          className={styles.SidebarLabel}
+          style={{ display: "grid", gridTemplateColumns: "1fr auto auto" }}
+        >
+          {"Obstacle x: " +
```
shueja, line 53: Could there be a better way to distinguish obstacles? The coordinates in the title feels kind of clunky.

src/document/CircularObstacleStore.tsx
```diff
@@ -0,0 +1,62 @@
...
+    },
+    setY(y: number) {
+      self.y = y;
+    },
+    setRadius(radius: number) {
+      self.radius = radius;
+    },
+    setSelected(selected: boolean) {
+      if (selected && !self.selected) {
+        const root = getRoot<IStateStore>(self);
```
shueja, line 51: If root is undefined, return early.

src/components/field/svg/FieldObstacles.tsx
```diff
@@ -0,0 +1,116 @@
...
+  appendIndexID(id: string): string {
+    return `${id}${this.props.index}`;
+  }
+
+  dragPointTranslate(event: any) {
+    let pointerPos = { x: 0, y: 0 };
+    pointerPos.x = event.x;
+    pointerPos.y = event.y;
+
+    this.props.obstacle.setX(pointerPos.x);
```
shueja, line 30: This snaps the obstacle center to the pointer, which does not work well on drag targets the size of the obstacle.

src/components/config/CircularObstacleConfigPanel.tsx
```diff
@@ -0,0 +1,61 @@
...
+              enabled={true}
+              setEnabled={(a) => null}
+              number={obstacle.y}
+              setNumber={(y) => obstacle!.setY(y)}
+              showCheckbox={false}
+              titleTooltip={"Obstacle Center Y"}
+            />
+
+            <Input
+              title="radius"
```
shueja, line 46: Consider leaving the title just `r` for visual lineup.

src/document/HolonomicPathStore.ts
```diff
@@ -20,9 +20,14 @@ import {
   WaypointID,
   WaypointScope,
 } from "./ConstraintStore";
-import { SavedWaypointId } from "./previousSpecs/v0_1";
+import { SavedWaypointId } from "./previousSpecs/v0_1_2";
```
shueja, line 23: This should actually be forward-exported from DocumentSpecTypes, but that's outside the scope of this PR

src/document/HolonomicPathStore.ts
```diff
@@ -77,6 +84,9 @@ export const HolonomicPathStore = types
       },
       asSavedPath(): SavedPath {
         let trajectory: Array<SavedTrajectorySample> = self.generated;
+        console.log(
```
shueja, line 87: To Remove

src/document/PathListStore.ts
```diff
@@ -3,6 +3,7 @@ import { SavedPathList } from "./DocumentSpecTypes";
 import { HolonomicPathStore } from "./HolonomicPathStore";
 import { v4 as uuidv4 } from "uuid";
 import { ConstraintStores } from "./ConstraintStore";
+import { CircularObstacleStore } from "./CircularObstacleStore";
```
shueja, line 6: Unnecessary import?

shueja (changes requested):
image
obstacles are on a z-index between trajectories/initial guess lines and constraint targets/waypoints. Intended?

Dragging an obstacle snaps the center of the obstacle to the mouse. This was acceptable with waypoints, since the drag targets were small. Not so here, especially with large obstacles that only need to move a small amount.

The amount of code duplication here is partially inevitable, but could show need for some helper functions that would work on both the obstacles array and the waypoints array.

The small drag handle for radius is annoying to work with. Could the entire outer border of the circle be a resize handle? The existing handler for the drag event already supports this, the target for clicking just needs to be bigger.

The bright green for selected obstacles is visually jarring. Consider using var(--select-yellow) on the border and leaving the internals red (perhaps a more transparent red).

Should document schema 0.1.2 use the `circleObstacles` key instead of `obstacles` to avoid an annoying rename in the auto-upgrader later?

shueja (approved):

shueja:
Approval is conditional on adjusting the dependency location of trajoptlib back to sleipnir's main now that https://github.com/SleipnirGroup/TrajoptLib/pull/69 merged. Everything else looks good.

PR #207, opened 2023-12-25
Vendor UI fonts

max-niederman:
Right now, Choreo relies on the system's available fonts, which is inconsistent and makes for a bad UI if users have bad default fonts installed. This PR vendors Roboto and Source Code Pro, which are the most-preferred free fonts in the font-family directives Choreo currently has.

It might also be good to use exclusively these two fonts for improved consistency on macOS and Windows.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,17 +1,19 @@
 {
   "name": "choreo",
-  "version": "0.3.0",
+  "version": "2024.0.4",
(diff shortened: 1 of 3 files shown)


Lewis-Seiden (approved):

shueja:
We vendored Roboto and Roboto Mono in the robot config PR.

PR #213, opened 2024-01-01
Docs Refresh

Awesomeplayer165:
See demo on github pages: https://awesomeplayer165.github.io/Choreo/

PR #216, opened 2024-01-02
Update to new docs

Awesomeplayer165:
Awaits #213


Awesomeplayer165:
sorry, should be opening from choreo's repo..

Issue #243, opened 2024-01-10
bug: origin point on 2024 field image is off


rgodha24:
<img width="747" alt="Screenshot 2024-01-09 at 6 22 00 PM" src="https://github.com/SleipnirGroup/Choreo/assets/90421212/d002d939-eb81-4d3e-95c4-ae8174d435ca">

I'm on MacOS x86, both with the official release build and a local version. I think it's a Mac only (eg. Safari webkit) issue because the (windows? not mac for sure) image in the CD post (duplicated below) is fine

![image](https://github.com/SleipnirGroup/Choreo/assets/90421212/706238f0-7508-48f1-bcab-452d3e148e72)

Awesomeplayer165:
I can confirm it happens on ARM64 and Rosetta 2

rgodha24:
it seems to be the image that's off, because the lines are still at 0,0

<img width="698" alt="Screenshot 2024-01-09 at 6 53 33 PM" src="https://github.com/SleipnirGroup/Choreo/assets/90421212/d8377d7d-100b-4dd7-bbb1-0ee9c6d63b82">

rgodha24:
i found a fix, working on the PR rn

shueja:
Closed by #246

PR #245, opened 2024-01-10
fix macos field image bug

rgodha24:
#243


rgodha24:
i added the svgr vite plugin so we can avoid xlinkHref which was the cause of the issue

shueja:
Fixed in a different way by #246

PR #252, opened 2024-01-11
Add MathJax rendering to MkDocs

calcmogul:
https://squidfunk.github.io/mkdocs-material/reference/math/#mathjax-mkdocsyml


calcmogul:
I generated the site locally, and underscores aren't rendering into subscripts. The rest is rendering though.

calcmogul:
Turns out GitHub requires escaping the underscores for it to render, but MkDocs requires _not_ escaping them.

PR #276, opened 2024-01-18
Clone split sections of trajectory before editing their timestamps

shueja:


jlmcmchl (approved):
Confirmed that this addresses the path rendering bug caused by modifying the trajectory timestamps in place

Lewis-Seiden (approved):

PR #286, opened 2024-01-23
Added duplication of paths

shueja:


Lewis-Seiden:
image
disambiguating names does not seem to work properly

src/document/PathListStore.ts
```diff
@@ -44,18 +44,22 @@ export const PathListStore = types
   })
   .actions((self) => {
     return {
-      setActivePathUUID(uuid: string) {
-        if (self.pathUUIDs.includes(uuid)) {
-          self.activePathUUID = uuid;
-        }
-      },
-      addPath(name: string, select: boolean = false): string {
+      disambiguateName(name: string) {
```
Lewis-Seiden, line 47: image
this does not appear to work properly

Lewis-Seiden (changes requested):

src/document/PathListStore.ts
shueja: Works as intended, but my intent is probably non-ideal. I wanted a difference between copying "NewPath" twice vs copying NewPath and then copying that copy.

Lewis-Seiden: i guess that makes sense, but it does feel a little weird and arbitrary to me without being told that thats the logic behind it

shueja: I think it's fine as-is, considering that most people won't be spam-clicking it.

Lewis-Seiden (approved):

PR #291, opened 2024-01-27
Update Field Width Constant

GalexY727:
Taken directly from cad at https://cad.onshape.com/documents/dcbe49ce579f6342435bc298/w/b93673f5b2ec9c9bdcfec487/e/6ecb2d6b7590f4d1c820d5e3
Is there a possibility that because the link above is the optimized version the old number was gotten from the original import and thus this change is wrong?


shueja:
I think this just wasn't rechecked for 2024.

GalexY727:
> I think this just wasn't rechecked for 2024.

It doesn't make sense why it would change so I get that... I wonder what they used that 0.01m of real estate for 🤭

shueja (approved):

Issue #254, opened 2024-01-11
Integrate docs navigation and TOC together in the sidebar


nlaverdure:
Provides a more seamless experience reading the docs, because you can more easily see where you are.

https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#navigation-integration

PR #298, opened 2024-01-28
Docs Enhancement: Hide TOC and navigation from Home Page

Awesomeplayer165:
Hides the TOC and navigation from `index.html` home page. This allows the 3 cards under `What is Choreo` header. Changes it from this:

image

to:

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,3 +1,15 @@
+---
+hide:
+  - toc
+  - navigation
+---
+
+!!! tip inline end
+
+    See the navigation links in the header or side-bar.
+
+    Click :octicons-three-bars-16: (top left) on mobile.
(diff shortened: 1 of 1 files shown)


nlaverdure:
Definitely looks good to me!!

(It's just a little funny that the page feels the need to give instructions on how to operate the navigation. It makes you ask why it isn't self evident.)

TOCs are still present on all the other pages right? I'd recommend trying out the option that combines the TOC so it's inline with the navigation. (See #254)

What do you think about moving the navigation header (Home | Usage | ChoreoLib | Contributing) into the nav sidebar as well?

Preference: I just think it's a better ux if the navigation is all in one place and behaves modally as little as possible.

PR #306, opened 2024-02-01
Add event marker support

shueja:


Awesomeplayer165 (commented):
I would suggest adding markers to the visibility toolbar on the side? it is getting a bit long however.

calcmogul:
The UI part worked for me at least. Took me a minute to realize I needed to click on the waypoint to add the marker. I assume this'll be addressed by docs later.

calcmogul (commented):
The UI part worked for me at least. Took me a minute to realize I needed to click on the waypoint to add the marker. I assume this'll be addressed by docs later.

calcmogul (approved):

PR #310, opened 2024-02-02
Docs/navigation

### LINKED ISSUE #254: Integrate docs navigation and TOC together in the sidebar

nlaverdure:
The theme of this PR is: 
Navigation should always feel and behave the same no matter where you are on the site

- Integrated the TOC so it's inline with the navigation. This closes #254 
    - This also supersedes #298  since the TOC is always integrated into the nav
- Removed the navigation tip from the main page
    - The _hope_ is that the navigation is now so clear, users don't need extra help to navigate the site
- Clean up some headers that showed up in the TOC/nav unexpectedly
- Flattened the nav hierarchy so that the Document Settings Panel, Robot Configuration, and File Export Settings pages are not under a subfolder and instead are seen right away on the Home page navigation.
- Made the nav area sticky

### DIFF
--- a/docs/contributing/contributing-guide.md
+++ b/docs/contributing/contributing-guide.md
@@ -2,13 +2,6 @@
 
 Thank you for your interest in contributing to this project! We appreciate your help in making it better.
 
-## Table of Contents
-- [Introduction](#introduction)
-- [Submitting Issues](#submitting-issues)
-- [Pull Requests](#pull-requests)
-- [Code Style](#code-style)
-- [License](#license)
-
 ## Introduction
 
 This document outlines the guidelines for contributing to this project. It covers submitting issues, creating pull requests, and maintaining code quality.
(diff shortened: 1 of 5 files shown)


Awesomeplayer165:
For documentation purposes, this is what it currently looks like:

nlaverdure:
I don't have strong feelings on the navigation tip and can be convinced that it belongs back in if anyone thinks it's important. Definitely don't want to get in a revert war about it!

Awesomeplayer165:
Sorry about not responding. It looks good to me!

calcmogul (approved):

Awesomeplayer165 (approved):

PR #315, opened 2024-02-06
Fix "Estimating MOI" figures and captions

calcmogul:
mkdocs has dedicated tags for making this look nicer. The main downside is GitHub preview no longer works.


nlaverdure (approved):
Built locally, looks good to me!

PR #348, opened 2024-02-26
Fix field length variable name

calcmogul:
It was called field width by mistake.


shueja (approved):

shueja:
It was a copyover from the gui code, where fieldWidth is the long, horizontal dimension and fieldHeight is the short, vertical dimension.

PR #352, opened 2024-02-28
Add zoom-in and zoom-out functionality

Awesomeplayer165:
zoomIn and zoomOut functionality with `control +` and `control -` for zooming into and out of the field.
Currently, zoomIn shortcut is not working yet.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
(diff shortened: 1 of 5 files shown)


calcmogul:
Turns out you have to use the names of the unmodified/unshifted keys in the shortcut, hence `ctrl+=` instead of `ctrl-+` with `splitKey` set to `+`.

Awesomeplayer165:
oh right, i overthought that!

calcmogul (approved):

PR #384, opened 2024-03-05
fix all trajectories being stale on first load of a file

shueja:


src/document/HolonomicPathStore.ts
```diff
@@ -707,6 +706,8 @@ export const HolonomicPathStore = types
           marker.command.fromSavedCommand(saved.command);
           this.addEventMarker(marker);
         });
+        self.setIsTrajectoryStale(savedPath.isTrajectoryStale ?? false);
+        // this needs to be last or populating other parts of the path will set it to false
```
calcmogul, line 710: Comments should go above the code they're referencing.

calcmogul (changes requested):

calcmogul (approved):

Issue #36, opened 2022-12-29
Add commandline interface


calcmogul:
In the spirit of one-click builds, I'd like the build task of a freshly cloned robot project to noninteractively generate trajectories and output them to the deploy folder for GradleRIO to deploy. A noninteractive commandline mode would facilitate this.

This makes tweaking waypoints before a match via the project JSON or via the UI less error-prone, because it ensures the generated outputs during a deploy are up to date.

PR #391, opened 2024-03-09
Basic CLI Interface

NerdyPi:
This is based off of issue #36 and branch `feature/cli-generate`, though it is not a complete implementation of what I imagine would be a more ideal final solution. Currently, this change just adds the ability to open a project file from the command line, and specify if all trajectories should be immediately generated and deployed upon startup. 

In my opinion, the generation feature would probably be better if it bypassed the GUI entirely. However, since all parsing of project configuration is done in the frontend, doing so would require either: A) duplicating this functionality in the rust backend and calling it only through the CLI option, or, B) refactoring the frontend to allow for the rust backend to handle all relevant parts of the Choreo project state.


calcmogul:
You can run `npm run fmtJs` to fix the format issues.

calcmogul:
How does one actually use these CLI arguments? I can't figure it out locally.

NerdyPi:
> How does one actually use these CLI arguments? I can't figure it out locally. Are they positional or keyword arguments?

First argument is always the path to the project file. `choreo /path/to/project.chor` or `choreo project.chor` will both work (with the latter using the location where choreo is executed as the reference directory). 

The auto-generation argument is supplied afterwards. `choreo project.chor -a` or `choreo project.chor --generate-all` will run the generation

calcmogul:
I agree it would be best if the CLI mode didn't open the GUI at all for auto-generation. If that's not possible, it should at least autoclose so the process is non-interactive.

calcmogul:
The error message for bad CLI arguments isn't being printed properly. Maybe non-tauri parsing is the way to go?
```
[tav@myriad ~]$ choreo -chor-file
thread 'main' panicked at src/main.rs:473:10:
error while running tauri application: Setup(SetupError(FailedToExecuteApi(ParseCliArguments("error: Found argument '-c' which wasn't expected, or isn't valid in this context\n\n\tIf you tried to supply `-c` as a value rather than a flag, use `-- -c`\n\nUSAGE:\n    choreo [OPTIONS] [chor-file]\n\nFor more information try --help\n"))))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

shueja:
Superseded by #602

Issue #401, opened 2024-03-19
Opening Git Shared Chor file causes all Traj Event Markers to be set to []


changsolomon8:
This might be a product of attempting to use Choreo in the context of a git repository with multiple individual computers changing paths (and then just checking in a **subset of Traj** files).

Currently, with our set up, if you open our shared .chor file, it immediately touches N different trajectory files and clears out the eventmakers to be [].  This causes headaches because we then have to manually stash the changes and only check in a subset of the Traj files that we want to change.  We intentionally do not want to regenerate some Trajectories.  

Is there a better way/recommended way to use .chor files in git repository? Or is the recommendation to just regenerate all the Trajectories on each commit type of workflow?

shueja:
Seems like there's two issues here:
1. Trajectories all export upon opening the file, when this is not desired
2. When exporting trajectories upon opening the file, the event marker array is empty when there are properly-generated event markers that should be there.
Is the second an accurate description of what you experience?

changsolomon8:
Yes, the second is an accurate description here. 

I also replied a little to Adam, but regarding automatically exporting upon opening file: 

> Yeah I’m not advocating for footgunning other teams. Just trying to understand if there something we can do here to make this easier to use.  I suspect that if we don't fix this we will end up shooting ourselves in the food because we accidently deploy eventmarkers as [].

shueja:
Fixed in v2024.2.2

PR #404, opened 2024-03-22
Add new schema to save generated target of event marker

shueja:
Also fix logic for stale trajectory marker.

Related to #401, should fix the event markers being zeroed out since we no longer rely on isTrajectoryStale to deserialize them.

### DIFF
--- a/src/document/HolonomicPathStore.ts
+++ b/src/document/HolonomicPathStore.ts
@@ -800,15 +800,17 @@ export const HolonomicPathStore = types
         () => {
           // Reaction needs the return value to change,
           // so we can't just access the values and do nothing with them
-          // does not need toJS to do a deep check on this, since it's just a boolean
-          return {waypoints: toJS(self.waypoints),
-          constraints: toJS(self.constraints),
-        guessing: self.usesControlIntervalGuessing,
-        obstacles: toJS(self.obstacles)
-      };
+          
+          return {
+            waypoints: toJS(self.waypoints),
+            constraints: toJS(self.constraints),
+            // does not need toJS to do a deep check on this, since it's just a boolean
+            guessing: self.usesControlIntervalGuessing,
+            obstacles: toJS(self.obstacles)
+          };
         },
         (value) => {
-            self.setIsTrajectoryStale(true);
+          self.setIsTrajectoryStale(true);
         }
       );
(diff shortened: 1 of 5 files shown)


src/document/HolonomicPathStore.ts
```diff
@@ -807,17 +798,19 @@ export const HolonomicPathStore = types
...
-          return true;
+          // Reaction needs the return value to change,
+          // so we can't just access the values and do nothing with them
+          
+          return {
+            waypoints: toJS(self.waypoints),
+            constraints: toJS(self.constraints),
+            // does not need toJS to do a deep check on this, since it's just a boolean
+            guessing: self.usesControlIntervalGuessing,
+            obstacles: toJS(self.obstacles)
```
Awesomeplayer165, line 810: should this be a type?

src/document/previousSpecs/v0_3_1.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export const SAVE_FILE_VERSION = "v0.3.1";
+export interface SavedWaypoint {
+  x: number;
+  y: number;
+  heading: number;
+  isInitialGuess: boolean;
+  translationConstrained: boolean;
+  headingConstrained: boolean;
+  controlIntervalCount: number; // positive-integer>
```
Awesomeplayer165, line 13: make comments more uniform
```suggestion
  controlIntervalCount: number; // positive-integer
```

src/document/previousSpecs/v0_3_1.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+export type SavedGeneratedWaypoint = SavedWaypoint & {
+  timestamp: number;
+  isStopPoint: boolean;
+};
+export interface SavedTrajectorySample {
+  timestamp: number; //positive
```
Awesomeplayer165, line 20:
```suggestion
  timestamp: number; // positive
```

Awesomeplayer165 (changes requested):
looks good. keep going!

src/document/HolonomicPathStore.ts
shueja: No, it's not needed since this structure only exists in one spot.

Awesomeplayer165 (approved):

PR #415, opened 2024-03-26
Fix issues causing multiple autosaves on generation

shueja:
These issues also caused file corruption or trajectoryStale=false not being saved.


Awesomeplayer165 (approved):
lgtm

calcmogul (approved):

PR #419, opened 2024-03-27
Feature: Path Gradient

Awesomeplayer165:
A selectable path gradient:
1. None (what we have been using so far)
2. Velocity (default, what's most intuitive)
3. Progress (color indicating how far into the trajectory. green=start)
4. Centripetal (
5. Acceleration
6. Dt


Awesomeplayer165:
image

Awesomeplayer165:
Added menu.

Gold872:
I think the gradient colors for velocity should be swapped, it makes more sense to have it be green when it's slower since green is generally an indicator for something being small

shueja:
> I think the gradient colors for velocity should be swapped, it makes more sense to have it be green when it's slower since green is generally an indicator for something being small
> 
> Having additional gradients for angular velocity/acceleration could also be helpful in some scenarios

Counterpoint: red means slow on GPS apps.

Awesomeplayer165:
Changes agreed upon from discord:
adding segments, removing centripetal, and adding a description

src/components/config/robotconfig/PathGradient.tsx
```diff
@@ -0,0 +1,104 @@
...
+    function: PathGradientFunctions.progress
+  },
+  Acceleration: {
+    name: "Acceleration",
+    description: "Faster robot acceleration is expressed as green.",
+    function: PathGradientFunctions.acceleration
+  },
+  IntervalDt: {
+    name: "IntervalDt",
+    description: "Distance between intervals. Shorter is expressed as green.",
```
shueja, line 101: “Shorter time difference between intervals is shown as green”

I would prefer “shown” instead of “expressed” for all of these.

shueja (changes requested):
One thing about tooltips is all I can find from my phone.

Awesomeplayer165:
One thing I found from my phone is the descriptions should be rendered as tooltips.

Awesomeplayer165:
Added path segment functionality. made menu and toolbar nicer to interact with.

Awesomeplayer165:
Make color generation work for any path length by oscilating of i between 0 and 1 scaled by 360. The first segment will always be red since 0 is red on hsl scale

shueja (approved):

Awesomeplayer165:
Add angular velocity gradient

src/components/config/robotconfig/PathGradient.tsx
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  static angularVelocity(
+    point: SavedTrajectorySample,
+    i: number,
+    arr: SavedTrajectorySample[],
+    documentModel: IStateStore
+  ) {
+    const colorValue = Math.abs(
+      Math.sin(point.angularVelocity / 0.00000000001) * 40
```
calcmogul, line 91: This seems like a hack.

Issue #381, opened 2024-03-05
"Point toward/away from point" constraint


Lewis-Seiden:
useful for intaking and scoring where you want one end of the robot to face a game piece/goal

shueja:
This would be extra nice when used with https://github.com/SleipnirGroup/Choreo/issues/297

calcmogul:
This constraint could be implemented like so:
```
heading = atan2(point.y - y, point.x - x)
tan(heading) = (point.y - y)/(point.x - x)
sin(heading)/cos(heading) = (point.y - y)/(point.x - x)
sin(heading) * (point.x - x) = cos(heading) * (point.y - y)
```

shueja:
Is there a formulation that could allow a particular error boundary around the desired heading? I expect that would make convergence easier and most uses probably don’t require exact pointing.

calcmogul:
I'm not sure how to correctly manipulate nonlinear inequalities, so you'd probably be stuck with the following:
```
atan2(point.y - y, point.x - x) < heading + ε
atan2(point.y - y, point.x - x) > heading - ε
```

PR #427, opened 2024-04-02
Add PointAt constraint

bruingineer:
#381
along with https://github.com/SleipnirGroup/TrajoptLib/pull/140

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -18,7 +18,7 @@ tauri-build = { version = "1.5.1", features = [] }
 tauri = { version = "1.6.0", features = [ "window-close", "window-set-title", "path-all", "dialog", "dialog-confirm", "dialog-save", "dialog-open", "dialog-ask", "fs-all", "shell-open", "devtools"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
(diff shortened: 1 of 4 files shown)


src-tauri/Cargo.toml
```diff
@@ -18,7 +18,7 @@ tauri-build = { version = "1.5.1", features = [] }
 tauri = { version = "1.6.0", features = [ "window-close", "window-set-title", "path-all", "dialog", "dialog-confirm", "dialog-save", "dialog-open", "dialog-ask", "fs-all", "shell-open", "devtools"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
-trajoptlib = { git = "https://github.com/SleipnirGroup/TrajoptLib.git", rev = "325ddaa659cd36db8a87da363b66f66837df3e8f", features = ["sleipnir"] }
+trajoptlib = { git = "https://github.com/bruingineer/TrajoptLib.git", rev = "b3ae9e4bde9c67436e005f7c0a5547d203fca1da", features = ["sleipnir"] }
```
spacey-sooty, line 21: This will need to be updated after merge of https://github.com/SleipnirGroup/TrajoptLib/pull/140

calcmogul:
I've confirmed this works. There's a lot of gotchas around when the constraint is valid to use tho, vs immediately making the trajectory infeasible.

calcmogul:
A requirement when using this seems to be to bracket the constraint with two translation waypoints, so the heading has time to transition between a point-to region and a pose waypoint after it.

calcmogul (approved):

PR #429, opened 2024-04-03
Unselect sidebar and navbar items with escape key

calcmogul:


Awesomeplayer165 (approved):
in the future, we need a better system for sidebar management in code than numbers and -1.

PR #435, opened 2024-04-04
Feature: Add Open File Shortcut

Awesomeplayer165:
Command + O and Control + O.

### DIFF
--- a/src/document/DocumentManager.ts
+++ b/src/document/DocumentManager.ts
@@ -242,15 +242,17 @@ export class DocumentManager {
     });
     hotkeys.unbind();
     hotkeys("escape", () => {
-      this.model
+      this.model;
       this.model.uiState.setSelectedSidebarItem(undefined);
       this.model.uiState.setSelectedNavbarItem(-1);
     });
     hotkeys("ctrl+o,command+o", () => {
-      dialog.confirm(
-        "You may lose unsaved or not generated changes. Continue?",
-        { title: "Choreo", type: "warning" }
-      ).then((proceed) => {
+      dialog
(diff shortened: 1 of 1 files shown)


src/document/DocumentManager.ts
```diff
@@ -245,6 +245,9 @@ export class DocumentManager {
       this.model.uiState.setSelectedSidebarItem(undefined);
       this.model.uiState.setSelectedNavbarItem(-1);
     });
+    hotkeys("ctrl+o,command+o", () => {
+      this.saveFileDialog();
```
calcmogul, line 249: This opens the save file dialog, not the open file dialog.

calcmogul:
```suggestion
      invoke("open_file_dialog");
```

Awesomeplayer165: yeah not sure if it is possible with the current setup. ill need to refactor things around from mobx context to documentManager.

src/document/DocumentManager.ts
```diff
@@ -242,9 +242,20 @@ export class DocumentManager {
     });
     hotkeys.unbind();
     hotkeys("escape", () => {
+      this.model;
```
calcmogul, line 245: This line seems unnecessary.

calcmogul (changes requested):

src/document/DocumentManager.ts
Awesomeplayer165: not sure why that was even added..

Awesomeplayer165: I don't see this on my branch actually..

calcmogul (approved):

PR #438, opened 2024-04-04
Add heading offset to point at constraint

jlmcmchl:
PR follows https://github.com/SleipnirGroup/TrajoptLib/pull/144 - simply wiring up the heading offset parameter to the frontend.

I have noticed that it's _critical_ for successful generation that the initial guess accounts for point at constraints. If someone else could assist in implementing that, this would be a more robust PR.


src-tauri/src/main.rs
```diff
@@ -227,6 +227,7 @@ enum Constraints {
         x: f64,
         y: f64,
         tolerance: f64,
+        offset: f64,
```
spacey-sooty, line 230: Should this be called heading_offset?

calcmogul: Yes.

shueja:
One issue with this feature is that trajectory mirroring will mess with most usecases (where you want a non-zero robot-relative heading vector to point at a particular point on the field).

PR #439, opened 2024-04-04
Solver status feedback

shueja:


calcmogul:
Don't forget to fix the warnings.

src-tauri/src/main.rs
```diff
@@ -447,11 +459,33 @@ async fn generate_trajectory(
         path_builder.sgmt_polygon_obstacle(0, wpt_cnt - 1, o.x, o.y, o.radius);
     }
     path_builder.set_drivetrain(&drivetrain);
-    path_builder.generate(true)
+    path_builder.generate(true, handle)
 }
 
+/**
+ * A OnceLock is a synchronization primitive that can be written to once. Used here to
```
calcmogul, line 466: This comment is incomplete.

calcmogul (changes requested):

src/components/field/svg/FieldGeneratedLines.tsx
```diff
@@ -41,8 +45,8 @@ class FieldPathLines extends Component<Props, State> {
           /
           {this.context.model.uiState.selectedPathGradient !=
             PathGradients.None.name &&
-            generated.length > 1 &&
-            generated.map((point, i, arr) => {
+            trajectory.length > 1 &&
+            trajectory.map((point, i, arr) => {
               if (i == arr.length - 1) {
                 return <></>;
               }
```
Awesomeplayer165, line 52: reverse `if (!pathGradient) {` and `if (i == arr.length - 1) {`. technically, accessing the second point is more expensive than checking if something exists.

Awesomeplayer165: we can exit much faster than acccessing something for no reason.

Awesomeplayer165 (changes requested):
small things. cant wait to release this tho

src-tauri/src/main.rs
```diff
@@ -269,14 +273,22 @@ async fn cancel() {
...
 #[allow(non_snake_case)]
 #[tauri::command]
 async fn generate_trajectory(
+    _app_handle: tauri::AppHandle,
     path: Vec<ChoreoWaypoint>,
     config: ChoreoRobotConfig,
     constraints: Vec<Constraints>,
     circleObstacles: Vec<CircleObstacle>,
     polygonObstacles: Vec<PolygonObstacle>,
+    handle: i64,
```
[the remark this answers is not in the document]
Awesomeplayer165: (The other parameters are fairly obvious, but `handle` is a bit confusing to newer people on team).

src/components/field/svg/FieldGeneratedLines.tsx
Awesomeplayer165: ACTUALLY. we are already comparing if != PathGradients.None; therefore, implicitly checking against `undefined`, so it wouldn't change midway through map and before it (would invalidate current render...)

src/components/field/svg/FieldGeneratedLines.tsx
```diff
@@ -13,13 +13,17 @@ class FieldPathLines extends Component<Props, State> {
...
-      (point) => {
-        generatedPathString += `${point.x},${point.y} `;
-      }
-    );
+    const trajectory = path.generating
+      ? path.generationProgress
+      : path.generated;
+    // preserve this so that it updates when mutating the in-progress trajectory in place
+    //eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const iteration = path.generationIterationNumber;
```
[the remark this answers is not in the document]
shueja: It's needed because we mutate the generationProgress array in place for performance reasons, so I needed to depend on something else that will change on progress updates. I chose the iteration number which (for now) doesn't necessarily match the solver, but is incremented on every progress update. Later we can get the iteration number from the solver and perhaps show the value somewhere, but it will still be usable for this purpose.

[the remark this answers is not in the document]
Awesomeplayer165: Can we use the convention `_` instead of iteration, since the field you are getting is fairly self-documenting..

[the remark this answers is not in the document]
Awesomeplayer165: Also, we should leave a comment to future developers that may try to remove it in the future.

src/components/field/svg/FieldGeneratedLines.tsx
```diff
@@ -13,45 +13,45 @@ class FieldPathLines extends Component<Props, State> {
...
-      (point) => {
-        generatedPathString += `${point.x},${point.y} `;
-      }
-    );
+    const trajectory = path.generating
+      ? path.generationProgress
+      : path.generated;
+    // preserve the acccess of generationIterationNumber
+    // to trigger rerenders when mutating the in-progress trajectory in place
+    //eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Awesomeplayer165, line 23:
```suggestion

```

Awesomeplayer165: why is this still here?

src/components/field/svg/FieldSamples.tsx
```diff
@@ -11,12 +11,19 @@ class FieldSamples extends Component<Props, State> {
...
-
+  LINE_LENGTH = 0.15;
   render() {
     const path = this.context.model.document.pathlist.activePath;
+    const trajectory = path.generating
+      ? path.generationProgress
+      : path.generated;
+    // preserve the acccess of generationIterationNumber
+    // to trigger rerenders when mutating the in-progress trajectory in place
+    //eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Awesomeplayer165, line 22:
```suggestion

```

Awesomeplayer165 (changes requested):
other small things.

src/components/field/svg/FieldGeneratedLines.tsx
shueja: Because otherwise eslint fails because we don't use this value

calcmogul (approved):

PR #448, opened 2024-04-10
Feature: Context Menu

### LINKED ISSUE #446: Context menu for adding waypoint type at mouse point
Im thinking of a context [menu (mui)](https://mui.com/material-ui/react-menu/). It should not replace the current waypoint type, rather just place a waypoint at the cursor. From MUI docs, it should mirror `IconMenu` with the icons and colors matching with the waypoint type.

I can experiment with other information on the menu, like obstacles (since constraints/scopes don't make much sense). If groups are implemented, will look like:

Awesomeplayer165:
Closes #446


Awesomeplayer165:
https://github.com/SleipnirGroup/Choreo/assets/70717139/6bb46add-e5d1-462e-bae9-15747bb6f2d0

src/components/field/svg/FieldOverlayRoot.tsx
```diff
@@ -99,6 +109,48 @@ class FieldOverlayRoot extends Component<Props, State> {
...
+          clientY: this.context.model.uiState.contextMenuMouseSelection[1]
+        }),
+        contextMenuWaypointType
+      );
+      this.context.model.uiState.setContextMenuMouseSelection(undefined);
+    }
+
+    // User selects a waypoint
+    else if (
+      this.context.model.uiState.contextMenuSelectedWaypoint !== undefined
```
shueja, line 143: Do we need this to be an else if?

src/document/UIStateStore.tsx
```diff
@@ -273,7 +273,11 @@ export const UIStateStore = types
       types.union(
         ...Object.keys(PathGradients).map((key) => types.literal(key))
       )
-    )
+    ),
```
shueja, line 276: This would be a good chance to make a sub-store ContextMenuStore, for organizational purposes. Not sure if I need that changed now though.

shueja (approved):

src/components/field/svg/FieldOverlayRoot.tsx
Awesomeplayer165: Why would it not? There could be future conditions we could add, but I don't want to fully short circuit that yet.

src/document/UIStateStore.tsx
Awesomeplayer165: I can put that as code base quality issue for future.

src/components/field/svg/FieldOverlayRoot.tsx
calcmogul: This check is redundant and just adds noise.

calcmogul (changes requested):

src/components/field/svg/FieldOverlayRoot.tsx
Awesomeplayer165: right sorry, was looking at the code from my phone.

calcmogul (approved):

PR #456, opened 2024-04-23
Event trigger

Unclesdad:
Makes some new methods that return triggers for when the robot hits some event marker-like point in its path. 
Four new methods: 

- trajTrigger(), which will return a trigger that fires when the robot is on a given trajectory,
- `trajTrigger("example").whileTrue().exampleCommand()` will run exampleCommand when the robot is currently following example.traj. When the robot moves to the next trajectory, exampleCommand will stop occurring.

- pointTrigger(), which will return a trigger that fires when the robot is on a given trajectory, after a certain amount of time has passed since the robot has been on said trajectory (similar to an event marker)
- `pointTrigger("example", 3).whileTrue().exampleCommand()` will run exampleCommand when it has been 3 seconds since the robot has started on example.traj. When the robot moves to the next trajectory, exampleCommand will stop occurring.

- spanTrigger(), which will return a trigger that fires when the robot has been on a given trajectory for a certain period of time (rising edge), but will stop firing after a second period of time (falling edge).
- `spanTrigger("example", 3, 4).whileTrue().exampleCommand()` will run exampleCommand when it has been 3 seconds since the robot has started on example.traj, but will then stop running exampleCommand when it has been past 4 seconds since the robot has started on example.traj.

 - spotTrigger(), which will return a trigger that fires FOR a certain period of time AFTER a certain period of time. 
 - `spanTrigger("example", 3, 1).whileTrue().exampleCommand()` will run exampleCommand starting at 3 seconds past the robot starting on example.traj, however will stop at 4 seconds past starting at example.traj. This does the same function as spanTrigger(), however instead of taking in a starting point and an ending point in time, it takes in a starting point and a duration.
 
In order to find the current trajectory that the robot is on, I made it so that the swerve commands will update a static variable to the trajectory they were given. It's not very elegant, however it works since there is only going to be one robot using a Choreo path at a time.


choreolib/src/main/java/com/choreo/lib/Choreo.java
```diff
@@ -42,13 +47,15 @@ public static ChoreoTrajectory getTrajectory(String trajName) {
     return loadFile(traj_file);
   }
 
+
+
```
spacey-sooty, line 51: remove

choreolib/src/main/java/com/choreo/lib/Choreo.java
```diff
@@ -205,4 +217,79 @@ public static ChoreoControlFunction choreoSwerveController(
...
+  }
+
+  /**
+   * Returns a Trigger which fires if the robot is currently on a given ChoreoTrajectory.
+   * 
+   * @param trajName The file name (without the .traj) of the given trajectory.
+   * @return A Trigger which activates if the robot is on the trajectory trajName.
+   */
+  public static Trigger trajTrigger(String trajName) {
+    return new Trigger(Choreo::onTrajectory);
```
spacey-sooty, line 232: This won't build
```suggestion
    return new Trigger(() -> Choreo::onTrajectory(trajName));
```

choreolib/src/main/java/com/choreo/lib/Choreo.java
```diff
@@ -205,4 +217,79 @@ public static ChoreoControlFunction choreoSwerveController(
...
+   * @param offset The time between when the inputted trajectory is started and when the event
+   * trigger should fire.
+   * @return A Trigger which activates when the robot hits an event marker.
+   */
+  public static Trigger pointTrigger(String trajName, double offset) {
+    boolean started = false;
+    var timer = new Timer();
+    return new Trigger(() -> {
+      if (onTrajectory(trajName) && !started) {
+        started = true;
```
spacey-sooty, line 249: this should be a field on the class not a local variable

choreolib/src/main/java/com/choreo/lib/Choreo.java
```diff
@@ -205,4 +217,79 @@ public static ChoreoControlFunction choreoSwerveController(
...
+   * trajName, in seconds, that the trigger should stop firing.
+   * fallingEdge should be a greater number than risingEdge, or the Trigger will not fire.
+   * @return A Trigger which activates if the robot is on the trajectory trajName.
+   */
+  public static Trigger spanTrigger(String trajName, double risingEdge, double fallingEdge) {
+    boolean started = false;
+    var timer = new Timer();
+    return new Trigger(() -> {
+      if (onTrajectory(trajName) && !started) {
+        started = true;
```
spacey-sooty, line 275: This should also be a field on the class

choreolib/src/main/java/com/choreo/lib/Choreo.java
```diff
@@ -205,4 +217,79 @@ public static ChoreoControlFunction choreoSwerveController(
...
+   * hits an event marker.
+   * 
+   * @param trajName The file name (without the .traj) of the given trajectory.
+   * @param offset The time in seconds between when the inputted trajectory is 
+   * started and when the event trigger should fire.
+   * @param length The duration of the trigger in seconds, from start to finish.
+   * @return A Trigger which activates if the robot is on the trajectory trajName.
+   */
+  public static Trigger spotTrigger(String trajName, double offset, double length) {
+    return event(trajname, offset, length + offset);
```
spacey-sooty, line 293:
```suggestion
    return event(trajName, offset, length + offset);
```

spacey-sooty (commented):
Also has some whitespace issues I believe, you should run the formatter

spacey-sooty:
We've been using a modified version of this for our offseason code, once I've fleshed out any issues I can find I'd be interested in pushing this through

PR #467, opened 2024-04-26
apply angle modulus to initial guess dtheta

bruingineer:
normalizes the range of dtheta to -pi,pi. Now, the number of guess points added due to a rotation from 5pi/6 to -5pi/6 is the same as a rotation from 0 to 2pi/6.


calcmogul:
Could you provide a before and after screenshot/video of a generated trajectory?

bruingineer:
This affects path segments around initial guess waypoints and translational waypoints since those headings are not 'fixed' and wrapped.

## Current Behavior
> Waypoints point in the same direction
image

> Waypoints headings cross the -pi,pi boundary
image


## Pull Request Behavior
> Waypoints point in the same direction
image

> Waypoints headings cross the -pi,pi boundary
image

calcmogul (approved):

PR #469, opened 2024-04-29
Added a triggers api for auto routines

oh-yes-0-fps:
Removed old choreo command api


oh-yes-0-fps:
This will have to wait for a stabilization of the frontend/app on how it emits event markers in the trajectory json

shueja:
This will remain unmerged until we (hopefully) release a final 2024 version of Choreo before the UI rewrite, since this is such a breaking change.

oh-yes-0-fps:
That makes complete sense, i still have to test it on a robot irl. I have made a python script to change old .traj files to the new format so devs can beta test stuff until ui catches up.

The changes to the traj spec include the name of the trajectory in the json and an alternate event type thats simpler than pp event triggers.

choreolib/src/main/java/com/choreo/lib/Choreo.java
```diff
@@ -2,42 +2,64 @@
...
 import java.util.function.Supplier;
 
 /** Utilities to load and follow ChoreoTrajectories */
-public class Choreo {
+public final class Choreo {
   private static final Gson gson = new Gson();
 
   /** Default constructor. */
-  public Choreo() {}
+  public Choreo() {
```
shueja, line 30: Why not just make the constructor private?

choreolib/src/main/java/com/choreo/lib/ChoreoAutoLoop.java
```diff
@@ -0,0 +1,84 @@
...
+    this.loop = loop;
+    this.name = name;
+  }
+
+  /**
+   * Creates a {@link Trigger} that is true while this autonomous loop is being polled.
+   *
+   * @return A {@link Trigger} that is true while this autonomous loop is being polled.
+   */
+  public Trigger enabled() {
```
shueja, line 48: So what happens if you want to stop and restart the autonomous routine without restarting code? The loop.enabled().onTrue() triggers won't fire the second time?

choreolib/src/main/java/com/choreo/lib/ChoreoAutoTrajectory.java
```diff
@@ -0,0 +1,428 @@
...
+    return new Trigger(
+        loop,
+        new BooleanSupplier() {
+          double lastTimestamp = timer.get();
+
+          public boolean getAsBoolean() {
+            double nowTimestamp = timer.get();
+            return lastTimestamp < nowTimestamp && nowTimestamp >= timeSinceStart;
+          }
+        });
```
shueja, line 254: I don't think this works as intended. `lastTimestamp < nowTimestamp` will always be true because getAsBoolean always happens after the creation of the BooleanSupplier.

choreolib/src/main/java/com/choreo/lib/trajectory/ChoreoTrajectory.java
```diff
@@ -172,6 +196,16 @@ public ChoreoTrajectory flipped() {
...
     }
-    return new ChoreoTrajectory(flippedStates);
+    return new ChoreoTrajectory(name, flippedStates, this.events);
+  }
+
+  /**
+   * Returns the event with the given name.
+   *
+   * @param eventName The name of the event.
+   * @return The event with the given name, or an empty optional if the event does not exist.
```
shueja, line 206: These docs are wrong

choreolib/src/main/java/com/choreo/lib/ChoreoAutoTrajectory.java
```diff
@@ -0,0 +1,428 @@
...
+      List<ChoreoTrajectory> trajectories,
+      Supplier<Pose2d> poseSupplier,
+      ChoreoControlFunction controller,
+      Consumer<ChassisSpeeds> outputChassisSpeeds,
+      BooleanSupplier mirrorTrajectory,
+      Optional<Consumer<ChoreoTrajectory>> trajLogger,
+      Subsystem driveSubsystem,
+      EventLoop loop) {
+    this.name = "Group " + name;
+    this.trajectories = trajectories;
```
shueja, line 92: Should probably be an immutable copy of the list

choreolib-example/src/main/java/frc/robot/Robot.java
```diff
@@ -0,0 +1,209 @@
...
+  static {
+    var table = NetworkTableInstance.getDefault().getTable("choreo");
+    desiredPoseEntry = table.getStructTopic("DesiredPoseEntry", Pose2d.struct).getEntry(new Pose2d());
+    currentPoseEntry = table.getStructTopic("CurrentPoseEntry", Pose2d.struct).getEntry(new Pose2d());
+    desiredSpeeds = table.getStructTopic("DesiredSpeeds", ChassisSpeeds.struct).getEntry(new ChassisSpeeds());
+    desiredPIDSpeeds = table.getStructTopic("DesiredPIDSpeeds", ChassisSpeeds.struct).getEntry(new ChassisSpeeds());
+    currentSpeeds = table.getStructTopic("CurrentSpeeds", ChassisSpeeds.struct).getEntry(new ChassisSpeeds());
+    timestamp = table.getDoubleTopic("Timestamp").getEntry(0);
+
+    desiredPoseEntry.set(desiredPoseEntry.get());
```
shueja, line 167: What's the point of this section?

oh-yes-0-fps: That was just some debug stuff trying to find a bug I made (I did timer reset instead of restart in the auto traj :/)

calcmogul:
The robot project files add a lot of bloat I'd like to avoid. The WPILib VS Code plugin supports vendordeps adding custom examples, so we should look into that.

oh-yes-0-fps:
I removed the example project for now as it should either be in its own repo or inside the vendor dep, [my offseason code](https://github.com/McQuaidRobotics/2024_Crescendo/blob/development/src/main/java/com/igknighters/commands/autos/AutoRoutines.java) can be used as a preliminary example of the trigger features

Issue #479, opened 2024-05-22
Expose module force vectors in .traj and ChoreoLib


calcmogul:
TrajoptLib already provides this information in the solution, so we just need to plumb it.

shueja:
This will also make https://github.com/SleipnirGroup/Choreo/issues/459 easier for showing plots not currently captured in the .traj.

shueja:
Closed in #493

PR #480, opened 2024-05-22
Add module forces to trajectory states

BBScholar:
Addressing: https://github.com/SleipnirGroup/Choreo/issues/479  

General Thoughts/Questions:
- Do I need to think about the number of modules not being equal to 4? 

Todo: 
- [x] Ensure Serialization/Deserialization works for ChoreoLib (C++ and Java)


choreolib/src/main/native/include/choreo/lib/ChoreoTrajectoryState.h
```diff
@@ -91,6 +99,12 @@ class ChoreoTrajectoryState {
   /// The angular component of the velocity at that point in the trajectory
   units::radians_per_second_t angularVelocity = 0_rad_per_s;
 
+  /// The forces on the modules in the X direction
+  std::array<units::newton_t, 4> moduleForcesX{{0_N, 0_N, 0_N, 0_N}};
+
+  /// The forces on the modules in the Y direction
+  std::array<units::newton_t, 4> moduleForcesY{{0_N, 0_N, 0_N, 0_N}};
```
spacey-sooty, line 106: Do we wanna use `wpi::SmallSet` or something similar here?

choreolib/src/main/native/cpp/choreo/lib/ChoreoTrajectoryState.cpp
```diff
@@ -49,14 +54,24 @@ frc::ChassisSpeeds ChoreoTrajectoryState::GetChassisSpeeds() const {
 ChoreoTrajectoryState ChoreoTrajectoryState::Interpolate(
     const ChoreoTrajectoryState& endValue, double i) const {
   frc::Pose2d interpPose = frc::Interpolate(GetPose(), endValue.GetPose(), i);
+
+  // TODO: Can this be done cleaner?
+  std::array<units::newton_t, 4> lerpFX, lerpFY;
+  for (int idx = 0; idx < 4; ++idx) {
+    lerpFX[idx] = wpi::Lerp(moduleForcesX[idx], endValue.moduleForcesX[idx], i);
+    lerpFY[idx] = wpi::Lerp(moduleForcesY[idx], endValue.moduleForcesY[idx], i);
+  }
```
spacey-sooty, line 63: If we store these using an interpolatable type it could be cleaner.

choreolib/src/main/java/com/choreo/lib/ChoreoTrajectoryState.java
```diff
@@ -91,14 +104,24 @@ public ChoreoTrajectoryState interpolate(ChoreoTrajectoryState endValue, double
     double scale = (t - this.timestamp) / (endValue.timestamp - this.timestamp);
     var interp_pose = getPose().interpolate(endValue.getPose(), scale);
 
+    // NOTE: Could maybe do this with streams? This seems more efficient for now.
+    double[] interp_fx = new double[4];
+    double[] interp_fy = new double[4];
+    for (int i = 0; i < 4; ++i) {
+      interp_fx[i] = MathUtil.interpolate(this.moduleForcesX[i], endValue.moduleForcesX[i], scale);
+      interp_fy[i] = MathUtil.interpolate(this.moduleForcesY[i], endValue.moduleForcesY[i], scale);
```
spacey-sooty, line 112: If streams or some form of map could avoid allocating two new Arrays here that would be nice

spacey-sooty (commented):
> Do I need to think about the number of modules not being equal to 4?
Does Trajoptlib/Choreo provide options to modify this? If so I would say yes

Does the C++ version compile? I've noticed a few places where units are passed into std functions which shouldn't work

jlmcmchl:
Can this PR include documentation of the order of modules in the array? It's not obvious from this PR which is which.

choreolib/src/main/java/com/choreo/lib/ChoreoTrajectoryState.java
calcmogul: Streams don't necessarily avoid allocation. An array of primitives should end up being stack-allocated.

choreolib/src/main/native/include/choreo/lib/ChoreoTrajectoryState.h
calcmogul: Why?

choreolib/src/main/native/cpp/choreo/lib/ChoreoTrajectoryState.cpp
calcmogul: They're already inherently interpolatable. Locality of behavior is more important than abstraction.

choreolib/src/main/java/com/choreo/lib/ChoreoTrajectoryState.java
BBScholar: This was my thinking. The overhead of streams isn't worth it here IMO, especially given how simple the logic is.

BBScholar:
> > Do I need to think about the number of modules not being equal to 4?
> 
> Does Trajoptlib/Choreo provide options to modify this? If so I would say yes
> 
> Does the C++ version compile? I've noticed a few places where units are passed into std functions which shouldn't work

I don't think Choreo itself supports more or less than 4 modules currently. Not sure about TrajoptLib, need to check.

This compiles on my machine (minus javadoc for some reason). Not sure why the units classes wouldn't work with standard lib. Are you talking about `std::transform` and similar?

BBScholar:
For the document upgrade logic. Is it valid to add the module force arrays (zero initialized) and mark the paths as stale?

choreolib/src/main/native/cpp/choreo/lib/ChoreoTrajectoryState.cpp
```diff
@@ -49,14 +53,24 @@ frc::ChassisSpeeds ChoreoTrajectoryState::GetChassisSpeeds() const {
...
       interpPose.Rotation().Radians(),
       wpi::Lerp(velocityX, endValue.velocityX, i),
       wpi::Lerp(velocityY, endValue.velocityY, i),
-      wpi::Lerp(angularVelocity, endValue.angularVelocity, i)};
+      wpi::Lerp(angularVelocity, endValue.angularVelocity, i),
+      lerpFX,
+      lerpFY};
 }
 
 std::array<double, 7> ChoreoTrajectoryState::AsArray() const {
```
BBScholar, line 76: Should this include module data now? Leaning towards no.

calcmogul: Idk what this function is even for, cuz we already support serialization a different way.

BBScholar: I'll just leave it be then.

spacey-sooty (commented):
Does as array need to be updated?

calcmogul:
See https://github.com/SleipnirGroup/Choreo/pull/480/files/f2e212b53e282ce44c9706eff79e14cbbf854439#r1612273741.

BBScholar:
Feels like that should be its own PR. But I can remove it here if that's best for you.

calcmogul:
https://github.com/SleipnirGroup/Choreo/pull/482

BBScholar:
Still need to check if (de)serialization works. Not sure if I should do that via unit tests or local test projects. Unit testing seems easier.

BBScholar:
Wrote unit tests for json (de)serialization. Looks good. 

Tested the document upgrade logic on an old `.chor` file I had locally, looks good.

Should be ready for full review now.

spacey-sooty (approved):

src-tauri/tauri.conf.json
```diff
@@ -60,7 +60,7 @@
       {
         "fullscreen": false,
         "resizable": true,
-        "title": "Choreo v2024.2.2",
+        "title": "Choreo v2024.3.0",
```
calcmogul, line 63: This is typically done in a separate PR before release instead.

calcmogul (changes requested):

calcmogul (approved):

calcmogul:
I'd expect them to be in the same coordinate system as the velocity.

PR #489, opened 2024-06-14
[choreo] Upgrade to Tauri v2

### LINKED ISSUE #453: Upgrade to Tauri 2.0
https://v2.tauri.app/

calcmogul:
Fixes #453.


calcmogul:
The app has the following issues on my Linux machine:

1. keyboard shortcuts don't work until the window is in focus (and sometimes not even then)
2. the file open dialog crashes if I press Escape

Issue 1 is happening for me with 2024.2.2 as well, so it's probably not Tauri v2 causing it. If someone confirms neither of these issues occurs on Windows, I'm OK with merging this PR.

calcmogul:
Issue 1 was reproduced on Windows but not issue 2.

calcmogul:
I reproduced it locally on Arch. I had to generate the trajectory a few times, then it froze right when it finished generating.

spacey-sooty:
Keybinds dont work on my ubuntu 24.04 system but everything else seems to work

calcmogul:
Yea, I've had mixed results with the keybinds on Arch too. I've also heard the Windows version crashes randomly. I just updated the PR to Tauri v2's official release, so we'll see if that changes anything.

oh-yes-0-fps:
If we get the app running properly in Tauri v2 would we ship that for 2025? I'm personally in favor of doing so along as we properly stress test it.

calcmogul:
I'd say the cutoff is late November, when teams usually stop beta testing stuff.

calcmogul:
In addition to the flaky keyboard shortcuts, the app also doesn't exit after clicking the window close button. The window closes, but the app continues to run.

calcmogul:
Turns out the exit lag was from my system writing a coredump file (it doesn't have anything helpful in it, btw). The actual behavior is the app crashing immediately on exit.

calcmogul:
## Minimum reproducible example for crash on exit

[choreo-tauri-v2-crash.zip](https://github.com/user-attachments/files/18257386/choreo-tauri-v2-crash.zip)

1. Run `run.sh`
2. Click either button of the dialog that opens to close it
3. Click the main window close button
4. The app will crash on exit

Any use of a dialog from TypeScript will cause the application to crash on exit. If one isn't opened, the app does not crash.

## Fixes for crash on exit

The plumbing may be worse, but we could try opening dialogs from Rust instead: https://v2.tauri.app/plugin/dialog/#rust

tauri-dialog transitively uses gtk-rs for GTK bindings, which is no longer be maintained: https://github.com/PolyMeilex/rfd/issues/227. As for other backend options:

> JS | 6995AM — 12/23/24, 10:42 PM
> rfd would use XDG Portals, but those don't support message dialogs apparently, which is 4 of our 5 dialogs

## Fixes for file open dialog crashing on escape press

This was reported in https://github.com/PolyMeilex/rfd/issues/219 and fixed upstream. The fix hasn't appeared in a release yet.

calcmogul:
Recent Tauri upgrades fixed the crash on exit and file open dialog crashing on escape press.

PR #497, opened 2024-06-19
Update Trajoptlib: fix use-after-free in solver status

shueja:


calcmogul (changes requested):
Cargo.lock needs to be updated as well.

calcmogul (approved):

PR #511, opened 2024-07-08
Import TrajoptLib from commit 3964cce

calcmogul:
We're merging the TrajoptLib source into Choreo because their APIs are very tightly coupled. This lets us catch breaking changes earlier and easier.


spacey-sooty (commented):
This should probably include a README update to explain the structure of the repo and link to subproject READMEs

Issue #211, opened 2023-12-31
Support differential drive constraints


calcmogul:
The tool currently only supports swerve drivetrain constraints. Differential drive is another popular drivetrain platform in FRC.

This will require corresponding support in TrajoptLib (new constraint wrapper for diff drive, among other things).

PR #525, opened 2024-07-11
[trajoptlib] Add differential drive support

spacey-sooty:
Resolves https://github.com/SleipnirGroup/Choreo/issues/211


trajoptlib/src/DifferentialTrajectoryGenerator.cpp
```diff
@@ -0,0 +1,132 @@
...
+         path.waypoints.at(sgmtIndex + 1).segmentConstraints) {
+      size_t startIndex = GetIndex(N, sgmtIndex + 1, 0);
+      size_t endIndex = GetIndex(N, sgmtIndex + 2, 0);
+
+      for (size_t index = startIndex; index < endIndex; ++index) {
+        Pose2v pose{
+            x.at(index), y.at(index), {thetacos.at(index), thetasin.at(index)}};
+        Translation2v linearVelocity{(vL.at(index) + vR.at(index)) / 2, 0.0};
+        auto angularVelocity =
+            (vL.at(index) - vR.at(index)) / path.drivetrain.trackwidth;
```
calcmogul, line 53: Should be vR - vL for CCW+ rotation.

trajoptlib/src/path/DifferentialPathBuilder.cpp
```diff
@@ -0,0 +1,214 @@
...
+  if (targetIndex > greatestIndex) {
+    for (int64_t i = greatestIndex + 1; i <= targetIndex; ++i) {
+      path.waypoints.emplace_back();
+      initialGuessPoints.emplace_back(std::vector<Pose2d>{{0.0, 0.0, {0.0}}});
+      if (i != 0) {
+        controlIntervalCounts.push_back(40);
+      }
+    }
+  }
+}
```
calcmogul, line 212: You can deduplicate the common stuff between swerve and differential with either inheritance or composition.

calcmogul (changes requested):
The problem setup, kinematics constraints, and dynamics constraints seem to be missing from the DifferentialTrajectoryGenerator constructor.

Some tests and examples are needed to confirm this works right, which would have likely caught that many constraints are missing.

trajoptlib/src/path/DifferentialPathBuilder.cpp
spacey-sooty: Should this be done for the functions inlined functions in SwerveTrajectoryGenerator and DifferentialTrajectoryGenerator

spacey-sooty:
The example is broken right now and needs to be fixed

trajoptlib/src/DifferentialTrajectoryGenerator.cpp
```diff
@@ -0,0 +1,251 @@
...
+        Pose2v pose{
+            x.at(index), y.at(index), {thetacos.at(index), thetasin.at(index)}};
+        Translation2v linearVelocity{(vL.at(index) + vR.at(index)) / 2, 0.0};
+        auto angularVelocity =
+            (vR.at(index) - vL.at(index)) / path.drivetrain.trackwidth;
+        Translation2v linearAcceleration{
+            ((vL.at(index) + vR.at(index)) / 2 -
+             (vL.at(index - 1) + vR.at(index - 1)) / 2) /
+                dt.at(index),
+            0.0};
```
calcmogul, line 58: This isn't necessarily zero. Consider centripetal acceleration.

spacey-sooty:
Examples no longer crash but fail with "Exit status: too few degrees of freedom"

calcmogul:
That exist status means you have more equality constraints than decision variables. Either some of them are redundant and can be removed, or some of them conflict and your problem is infeasible.

trajoptlib/src/DifferentialTrajectoryGenerator.cpp
```diff
@@ -0,0 +1,273 @@
...
+      size_t index = GetIndex(N, wptIndex + 1, 0) - 1;
+      Pose2v pose{
+          x.at(index), y.at(index), {thetacos.at(index), thetasin.at(index)}};
+      Translation2v linearVelocity{(vL.at(index) + vR.at(index)) / 2, 0.0};
+      auto angularVelocity =
+          (vR.at(index) - vL.at(index)) / path.drivetrain.trackwidth;
+      Translation2v linearAcceleration{
+          ((vL.at(index) + vR.at(index)) / 2 -
+           (((index > 0) ? vL.at(index - 1) : 0.0) +
+            ((index > 0) ? vR.at(index - 1) : 0.0)) /
```
calcmogul, line 145: An if statement would be easier to read. Also, you shouldn't need to use finite backward difference here because linear and angular acceleration should be decision variables (at least, they were for swerve).

calcmogul (changes requested):

calcmogul:
I got the optimizer to produce solutions, though I haven't plotted them to make sure they make sense.

calcmogul:
When writing the Rust FFI, you can copy-paste all the swerve ones except point_at(). That one isn't feasible unless the drivetrain is moving toward or away from the point, so it's not useful.

spacey-sooty:
I'll work on Rust FFI now

calcmogul:
Actually, we can still include wpt_point_at(). It's sgmt_point_at() that has the feasibility issues.

trajoptlib/include/trajopt/trajectory/DifferentialTrajcectory.hpp
jlbabilino, line 1: spelling error on file name

Issue #341, opened 2024-02-22
Replace the dot handle on the robot UI element with a triangle


calcmogul:
A triangle makes it clearer whether that handle is attached to the front or back of the robot.

Nsl106:
Not sure if this issue is even still something that needs to be done, but figured I would find something easy to try to make sure I set everything up correctly :)
![image](https://github.com/SleipnirGroup/Choreo/assets/87742096/8313d2f9-c9e3-48a2-8fa5-275f67db9b0c)

PR #527, opened 2024-07-11
Replace circle heading handle with triangle

Nsl106:
Intended to fix https://github.com/SleipnirGroup/Choreo/issues/341, just a simple change to make the heading handle on waypoints a triangle so the front of the robot is more clear.
347597606-8313d2f9-c9e3-48a2-8fa5-275f67db9b0c


calcmogul:
The clickable area feels too small with a side length of 0.2. A side length of 0.25 gives equivalent area:
```
circle area = triangle area
πr² = 1/2 l²
2πr² = l²
l = √(2πr²)
l = r√(2π)
l = 0.1 √(2π)
l = 0.25066
l ≈ 0.25
```

calcmogul (changes requested):
The clickable area feels too small with a side length of 0.2. A side length of 0.26935 gives equivalent area:
```
triangle area = circle area
1/2 bh = πr²
1/2 l √3/2 l = πr²
√3/4 l² = πr²
l² = 4πr²/√3
l = √(4πr²/√3)
l = 2r√(π/√3)
l = 2 * 0.1 √(π/√3)
l ≈ 0.26935
```

calcmogul (changes requested):
The triangle point math in render() checks out.

The white circle on the playback robot needs to be changed to a triangle. I think it's rendered in InterpolatedRobot.tsx.
Screenshot_20240711_124026

shueja:
> The triangle point math in render() checks out.
> 
> The white circle on the playback robot needs to be changed to a triangle, but idk where that's rendered. Screenshot_20240711_124026

shueja:
The playback robot is InterpolatedRobot.tsx.

Nsl106:
Where should the shared code between them go? Or would it be better to just copy-paste the same math from OverlayWaypoint.tsx? Also thanks for cleaning up the math comments :)

calcmogul:
It's not that much code, so I think it's fine to copy-paste.

Nsl106:
What would you think about making the center circle a triangle or arrow (similar to the advantagescope odometry page), and making the heading handle only be visible when the waypoint is selected?

calcmogul:
> making the heading handle only be visible when the waypoint is selected

That makes it less clear how you're supposed to interact with a pose waypoint.

Nsl106:
It could just be if you grab it anywhere but the heading handle, not sure if that would be clear enough though.

calcmogul:
Having the waypoint translate if you missed the rotation handle sounds frustrating.

calcmogul (approved):

PR #542, opened 2024-07-21
[trajoptlib] Use heading for tank

spacey-sooty:
I couldn't make this build without splitting the headers


calcmogul:
Why? They're vectors of doubles.

calcmogul:
The differential example reports too few degrees of freedom.

spacey-sooty:
I got straight line paths working with this (they didn't work on main)

calcmogul (approved):

PR #552, opened 2024-08-09
Exclude project contents from LocalStorage when saving last opened file path

### LINKED ISSUE #551: Project Contents should be excluded from LocalStorage
The LocalStorage api has a hard limit of 5MB per item, which is easy to surpass with the full contents of a `.chor` file. 

The `Quota Exceeded` exception is thrown at this line: https://vscode.dev/github/SleipnirGroup/Choreo/blob/main/src/document/DocumentManager.ts#L114

To fix this, `payload.contents` should be removed from local storage. It appears that nothing needs to be done to address startup, as `openLastFile()` still calls the backend to load the file contents.

https://vscode.dev/github.com/SleipnirGroup/Choreo/blob/main/src/document/DocumentManager.ts#L62

bruingineer:
Fixes #551

### DIFF
--- a/src/document/DocumentManager.ts
+++ b/src/document/DocumentManager.ts
@@ -110,7 +110,7 @@ export class DocumentManager {
           this.model.uiState.setSaveFileName(saveName);
           this.model.uiState.setSaveFileDir(saveDir);
           this.model.uiState.setIsGradleProject(adjacent_gradle);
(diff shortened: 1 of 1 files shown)


shueja (approved):
Temporary fix (since document logic will change anyway) but does the job

PR #560, opened 2024-08-22
[choreo] use maximum angular velocity in guessControlIntervalCount

bruingineer:
Gives better interval counts for paths with slow angular velocity constraints


bruingineer:
this PR is for reference-only while a new pr will be submitted for the 2025 rearch.

PR #588, opened 2024-08-27
Add UI elements that reflect constraints

shueja:
Also, performance fixes:
- make serialize calls computed/cached properties
- when directly setting a default unit value on an expression store, avoid parsing the new expression when we already know the default magnitude
Known weaknesses:
Does not do much for tightly overlapping waypoints re: adding constraints. Consider changes to the sidebar to highlight/give constraint add targets in that list.


src/components/field/svg/constraintDisplay/FieldConstraintAddLayer.tsx
```diff
@@ -0,0 +1,255 @@
...
+    //                     });
+    //                     doc.history.stopGroup();
+    //                   }}
+    //                 ></circle>
+    //               </Fragment>
+    //             );
+    //           }
+    //         })}
+    //     </>
+    //   );
```
Lewis-Seiden, line 251: should this be removed?

src/document/UIStateStore.tsx
```diff
@@ -20,6 +20,12 @@ import {
 
 export const UIStateStore = types
   .model("UIStateStore", {
+    fieldMatrix: types.frozen<DOMMatrix>(
```
Lewis-Seiden, line 23: document this or FieldCTM somewhere, or at least expand out the acryonym

Lewis-Seiden (changes requested):
couple documentation things

Lewis-Seiden (approved):

PR #590, opened 2024-08-27
[rust] Added strict linting, formatting and better error types

oh-yes-0-fps:
I will expand more into the organization and diagnostics of the rust backend code, but for now this is a good start


rustfmt.toml
```diff
@@ -0,0 +1,29 @@
...
+control_brace_style = "AlwaysSameLine"
+trailing_semicolon = true
+trailing_comma = "Vertical"
+match_block_trailing_comma = false
+merge_derives = true
+use_try_shorthand = true
+use_field_init_shorthand = true
+force_explicit_abi = true
+condense_wildcard_suffixes = true
+edition = "2021"
```
spacey-sooty, line 29: Why set so many settings?

spacey-sooty (changes requested):
Linting and formatting should be checked in CI

calcmogul:
It _is_ checked in CI:
https://github.com/SleipnirGroup/Choreo/blob/main/.github/workflows/lint-format.yml#L54
https://github.com/SleipnirGroup/Choreo/blob/main/package.json#L16-L22

rustfmt.toml
oh-yes-0-fps: Most of them are default but incase somebody has a local user wide fmt config this will override that

src-tauri/Cargo.toml
```diff
@@ -23,6 +23,7 @@ open = "3"
 lazy_static = "1.4.0"
 tokio = { version = "1", features = ["full"] }
 dashmap = "6.0.1"
+thiserror = "1.0.63"
```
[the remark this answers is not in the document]
oh-yes-0-fps: I disagree that `thiserror` is only for libraries, it's good at organizing variants of errors especially if you want to document emittable variants. Also most of this code is still being consumed by other code in the context of the frontend so it's kinda like a library.

rustfmt.toml
spacey-sooty: Maybe add a comment for which ones aren't the default and why you reset the defaults

src-tauri/Cargo.toml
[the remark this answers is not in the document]
spacey-sooty: Ehhh I see the argument for it, either way I think Trajoptlib should have better errors using `thiserror` (then TrajoptError can wrap that)

rustfmt.toml
oh-yes-0-fps: Actually, I didn't realize how many of those are unstable because I'm always on nightly, our ci can't utilize the nightly ones so ill trim the toml down alot

src-tauri/Cargo.toml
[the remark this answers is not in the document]
oh-yes-0-fps: I agree, I think that is slightly outside the scope of this pr but I'll make an issue for it and will get back to it later.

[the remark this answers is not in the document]
spacey-sooty: I guess I'll do it Im working on Trajoptlib rn anyways

src-tauri/src/document/types.rs
```diff
@@ -288,34 +295,39 @@ pub struct Sample {
...
+
 #[derive(Serialize, Deserialize, Clone)]
+#[allow(clippy::struct_field_names)]
 pub struct Traj {
     pub name: String,
     pub version: String,
     pub path: ChoreoPath<Expr>,
     // Captures the path that formed the current generated trajectory
     pub snapshot: Option<ChoreoPath<f64>>,
+    // TODO: maybe rename to `output`, this is a breaking change for frontend though
```
spacey-sooty, line 331: Now is a good time to be breaking things...

oh-yes-0-fps: i wanted this to be a purely non functional pr so it can get merged quickly

shueja (approved):

PR #601, opened 2024-08-28
[trajoptlib] Use thiserror for error types

spacey-sooty:


oh-yes-0-fps:
What is this waiting on or what still needs to be done?

spacey-sooty:
Fixing conflicts, properly propogating the error from C++ -> Rust and API design decision. I'm gonna take another look at this in a bit to fix some of this stuff up

trajoptlib/src/error.rs
```diff
@@ -0,0 +1,47 @@
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum TrajoptError {
+    #[error("The solver determined the problem to be overconstrained and gave up.")]
```
calcmogul, line 5: The fact this duplicates the contents of https://github.com/SleipnirGroup/Sleipnir/blob/main/include/sleipnir/optimization/SolverExitCondition.hpp#L47-L78 isn't ideal.

spacey-sooty: yeah, do you have a better option?

calcmogul: See my edited message.

oh-yes-0-fps (approved):

Issue #581, opened 2024-08-26
Add remaining constraints back to 2025 ui


shueja:
Except for Waypoint Velocity Direction, and Straight Line (see #580)

PR #603, opened 2024-08-28
[choreo] add angular velocity constraint

bruingineer:
add angular velocity constraint to the ui
use the constraint to get a better control interval guess

part of #581


src-tauri/src/document/intervals.rs
```diff
@@ -58,17 +58,28 @@ pub fn guess_control_interval_count(
...
+            let max_wheel_position_radius = config
+                .modules
+                .iter()
+                .fold(0f64, |max, &module| max.max(module.x.hypot(module.y)));
+            let max_ang_vel = max_vel / max_wheel_position_radius;
+
+            // use peak angular velocity and dtheta to calculate the time
+            // for theta to travel a cubic hermite spline profile
+            // in which the max angular velocity is not exceeded as an estimate
+            let time = (1.5 * dtheta) / max_ang_vel;
```
shueja, line 79: Where does this 1.5 number come from?

shueja (changes requested):
Unsure about the derivation of the 1.5 number. Otherwise lgtm

src-tauri/src/document/intervals.rs
bruingineer: It estimates the time needed to rotate `dtheta` without exceeding `max_ang_vel`. It comes from the cubic Hermite spline and moving around its variables.

/*
 * Proof for T = 1.5 * θ / ω:
 *
 * - The velocity function derived from the cubic Hermite spline is:
 *   v(t) = -6*θ*t^2 + 6*θ*t.
 *
 * - The peak velocity occurs at t = 0.5, where t∈[0, 1] :
 *   v(0.5) = 1.5*θ, which is the max angular velocity during the motion.
 * - To ensure this peak velocity does not exceed ω, we set:
 *   1.5 * θ = ω.
 * - The total time T needed to reach the final θ and
 *   not exceed ω is thus derived as:
 *   T = θ / (ω / 1.5) = 1.5 * θ / ω.
 *
 * This calculation ensures the peak velocity meets but does not exceed ω,
 * extending the time proportionally to meet this
 * requirement.
 */

calcmogul: That proof should be included in the source code for context.

shueja (approved):

PR #617, opened 2024-08-30
Fix placement of dotted line shown when placing constraints

### LINKED ISSUE #608: [choreo] The dotted line shown when placing constraints is totally out of place
Steps to reproduce:
1. Place down a waypoint.
2. Select a constraint and click on the waypoint.
3. Observe as the dotted line points off to infinity.

Choreo version: f571654eb2f043b85c86264485eea35ad3ba4246

shueja:
Reporter of #608 said this build fixed the issue.

Closes #608


shueja:
Adding an indent layer in FieldOverlayRoot kind of obscured the change here. The fix is to the field-to-screen matrix (used to transform mouse event coordinates to field coordinates). Putting it in uiState was causing some issues with missed resize events. By providing it as a context to any elements within FieldOverlayRoot (the only elements where it's useful anyway), we fix the issue according to the reporter.

oh-yes-0-fps (approved):

PR #618, opened 2024-08-31
[trajoptlib] Set better initial value of dt_sgmt variable for swerve

bruingineer:


shueja (approved):

spacey-sooty (commented):
Could a similar change be applied to DifferentialTrajectoryGenerator?

PR #619, opened 2024-08-31
[choreo] short circuit add constraint layer for constraints with only waypoint scopes

bruingineer:
stop points (and future waypoint-only constraints) should only be added to a single waypoint. this adjusts the constraint adding UI so that adding a stop point is completed after clicking a single waypoint.


shueja (approved):

shueja:
Another related issue is segment constraints that point to the same waypoint on both ends. But I think we already handle these.

PR #620, opened 2024-08-31
[ci] Shorten recursive wildcard

Gold856:
Tauri always puts the installer in the same location everytime. Removing the wildcard means the zip file can contain just the installer with no extra directories.

### DIFF
--- a/.github/workflows/choreo.yml
+++ b/.github/workflows/choreo.yml
@@ -82,14 +82,14 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: ${{ matrix.artifact-name }}
-          path: target/**/release/bundle/nsis/*.exe
+          path: "**/*.exe"
 
       - name: Upload bundle (macOS)
         if: startsWith(matrix.os, 'macOS')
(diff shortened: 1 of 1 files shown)


calcmogul (changes requested):
Tauri doesn't actually put the binary in the same place for all targets:
Screenshot_20240830_214756
https://github.com/SleipnirGroup/Choreo/actions/runs/10642482997?pr=620

.github/workflows/choreo.yml
```diff
@@ -82,23 +82,23 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: ${{ matrix.artifact-name }}
-          path: target/**/release/bundle/nsis/*.exe
+          path: "**/*.exe"
```
calcmogul, line 85: Are the quotes needed?

Gold856: Not sure, but my editor's syntax highlighting suggested that it needed quotes.

calcmogul (approved):

PR #623, opened 2024-08-31
[trajoptlib] Add aL and aR to DifferentialSolution

spacey-sooty:


calcmogul:
Why doesn't swerve also include it?

spacey-sooty:
> Why doesn't SwerveSolution also include acceleration?

It already does

calcmogul:
Well, SwerveTrajectorySample doesn't.

calcmogul (approved):

PR #594, opened 2024-08-28
[trajoptlib] Build: add CMakePresets.json for default CMake configs

spacey-sooty:
Also add CMakeUserPresets.json to .gitignore

PR #634, opened 2024-09-02
[ci] Fix TrajoptLib tests failing on Windows

### LINKED ISSUE #512: Windows CI can't run tests because it can't find DLLs
We could temporarily fix CI by copying the TrajoptLib, Sleipnir, and Catch2 DLLs to the executable directory, but that doesn't fix local builds.

Gold856:
Adds Sleipnir's build directory to the DLL search paths. [The Catch2 feature](https://github.com/catchorg/Catch2/blob/devel/extras/Catch.cmake#L120) that enabled this fix requires CMake >=3.22 though. I also modified one of tests that was failing on my system.

Fixes #512.

### DIFF
--- a/.github/workflows/trajoptlib-cpp.yml
+++ b/.github/workflows/trajoptlib-cpp.yml
@@ -14,11 +14,9 @@ jobs:
       matrix:
         include:
           - artifact-name: Windows x86_64
-            # FIXME: Tests give "Exit code 0xc0000135" for missing DLLs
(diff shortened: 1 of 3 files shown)


calcmogul:
CMake presets are broken on Windows: https://github.com/SleipnirGroup/Choreo/pull/594

Gold856:
Yeah, I'm going to keep the separate Windows setup because cross-compiling is painful otherwise.

calcmogul (approved):

PR #639, opened 2024-09-02
Fix CLI build on Windows

Gold856:


calcmogul:
I'm concerned that CI didn't catch this.

Gold856:
I get this error on main:
```
'cp' is not recognized as an internal or external command,
operable program or batch file.
```
That indicates that it's running using Command Prompt as the shell, not PowerShell. Why my laptop acts that way and not CI is a mystery to me.

calcmogul (approved):

PR #663, opened 2024-09-03
[trajoptlib] Fix Differential linear velocity and acceleration constraints

spacey-sooty:


trajoptlib/include/trajopt/DifferentialTrajectoryGenerator.hpp
```diff
@@ -13,6 +13,7 @@
 #include <sleipnir/optimization/OptimizationProblem.hpp>
 #include <sleipnir/optimization/SolverExitCondition.hpp>
 
+#include "trajopt/geometry/Translation2.hpp"
```
calcmogul, line 16: This is only used in the .cpp file.
```suggestion
```

calcmogul (changes requested):

calcmogul (approved):

PR #677, opened 2024-09-03
[choreo] Generation re-architecture

oh-yes-0-fps:
This pr rearchitects the generation function to be a series of transformations on the trajopt problem builder. This allows better implementation of beta features, less fear of merge conflicts for new generation features, easier declaring explicit transformations for diffy or swerve, and easier to maintain code by spreading things out.

[transformer usage](https://github.com/oh-yes-0-fps/Choreo/blob/generation-rearch/src-core/src/generation/generate.rs#L145)
[implementation of a transformer](https://github.com/oh-yes-0-fps/Choreo/blob/generation-rearch/src-core/src/generation/transformers/drivetrain_and_bumpers.rs)

Every transformer lays behind a feature string, the frontend can send in a list of features it wants active for that specific generation. The feature string of `""` implies the transformer is always active.


shueja (approved):
LGTM, thanks for the work!

PR #680, opened 2024-09-03
[choreo] Move logs to dirs::state_dir

calcmogul:
Tauri v2's log location getter now requires an app handle, which isn't available by the time setup_tracing() is called. Instead, this PR just uses the dirs package.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -386,6 +386,7 @@ dependencies = [
  "built",
  "choreo-core",
  "current_platform",
+ "dirs",
  "nu-ansi-term 0.50.1",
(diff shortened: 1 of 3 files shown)


spacey-sooty:
> Tauri v2's log location getter now requires an app handle, which isn't available by the time setup_tracing() is called. Instead, this PR just uses the dirs package.

Why not call setup_tracing in `setup()`?

calcmogul:
Setting up tracing and logs should be the first thing the program does, and setup() isn't the first thing the app runs.

oh-yes-0-fps (approved):

PR #687, opened 2024-09-04
[choreo] revert making intervals an option

bruingineer:


src-core/src/spec/traj.rs
```diff
@@ -20,13 +20,15 @@ pub struct Waypoint<T: SnapshottableType> {
...
     /// The number of control intervals to use between this waypoint and the next.
-    pub intervals: Option<usize>,
+    pub intervals: usize,
     /// Whether to split the trajectory at this waypoint.
     pub split: bool,
     /// TODO
     pub fix_translation: bool,
     /// TODO
     pub fix_heading: bool,
+    /// Whether to override the intervals.
```
shueja, line 30:
```suggestion
    /// Whether to override the intervals. Not an Option because unused overrides still get persisted to file.
```

shueja (changes requested):

shueja (approved):

PR #695, opened 2024-09-04
[choreo] Remove preprocess interval guessing

shueja:
If this works we can remove part of the tauri command api as well.


oh-yes-0-fps:
big fan

bruingineer:
I think this works. the counts do not update on the UI now. It would be helpful if the counts still updated on the UI since that informs any manually set override counts.

shueja:
A way to fix this is to send the interval counts back over the IPC after they're calculated.

PR #704, opened 2024-09-05
[trajoptlib] Remove an unwrap in generate

oh-yes-0-fps:


trajoptlib/src/lib.rs
```diff
@@ -683,7 +687,11 @@ impl DifferentialPathBuilder {
     ) -> Result<DifferentialTrajectory, TrajoptError> {
         match self.path_builder.generate(diagnostics, handle) {
             Ok(traj) => Ok(traj),
-            Err(msg) => Err(TrajoptError::from(msg.what().parse::<i8>().unwrap())),
+            Err(msg) => {
+                let what = msg.what();
+                println!("Trajoptlib Error: {}", what);
+                Err(TrajoptError::from(what.parse::<i8>().unwrap_or(-128)))
```
calcmogul, line 693: Does the parse failure mainly happen if the wrong type got thrown by C++?

trajoptlib/src/error.rs
```diff
@@ -19,6 +19,8 @@ pub enum TrajoptError {
     MaxIterationsExceeded,
     #[error("The solver returned its solution so far after exceeding the maximum elapsed wall clock time.")]
     Timeout,
+    #[error("The solver returned an unparsable error code")]
```
calcmogul, line 22: May as well be consistent with the other messages.
```suggestion
    #[error("The solver returned an unparsable error code.")]
```

calcmogul (changes requested):

trajoptlib/src/lib.rs
oh-yes-0-fps: Thats exclusively when it happens, i prolly coulda done a
`map_err(|_| TrajoptError::Unparsable)?` instead of `unwrap_or(-128)`

calcmogul (approved):

PR #721, opened 2024-09-09
[choreo] Frontend typing cleanup

oh-yes-0-fps:
This probably isnt done but i don't want to do everything in 1 pr.

I want the way everything fits together to be more apparent and reduce the surface area that the ui can use.


src/components/config/ConstraintsConfigPanel.tsx
```diff
@@ -1,6 +1,6 @@
 import { observer } from "mobx-react";
 import { Component } from "react";
-import { IConstraintStore } from "../../document/ConstraintStore";
+import { IConstraintStore } from "../../document/stores//ConstraintStore";
```
shueja, line 3: extra slash here, does that even work?

src/components/field/Field.tsx
```diff
@@ -12,11 +12,11 @@ import { CircularProgress, Tooltip } from "@mui/material";
 import Box from "@mui/material/Box/Box";
 import IconButton from "@mui/material/IconButton";
 import "react-toastify/dist/ReactToastify.min.css";
-import { ICircularObstacleStore } from "../../document/CircularObstacleStore";
-import { IConstraintStore } from "../../document/ConstraintStore";
-import { IEventMarkerStore } from "../../document/EventMarkerStore";
-import { IHolonomicWaypointStore } from "../../document/HolonomicWaypointStore";
-import { Commands } from "../../document/tauriCommands";
+import { ICircularObstacleStore } from "../../document/stores/CircularObstacleStore";
+import { IConstraintStore } from "../../document/stores//ConstraintStore";
```
shueja, line 16: Extra slash

shueja (changes requested):
Fix the `//` in the commented places and wherever else (seems to be just ConstraintStore)
Otherwise LGTM

PR #723, opened 2024-09-09
[choreo] Use new path name when duplicating

### LINKED ISSUE #722: [choreo] duplicate path button overwrites existing path
after clicking duplicate, the original path stays in the UI but will not be editable. The new path entry will be editable and overwrite the original .traj file.

bruingineer:
fixes #722


calcmogul (approved):
Works locally. For a path named "New Path", the duplicate is named "New Path (1)".

PR #726, opened 2024-09-09
[choreo] use finally instead of then

### LINKED ISSUE #725: [choreo] Cannot delete a path without a corresponding .traj file
paths in the sidebar without a .traj file cannot be deleted.  an app restart will get rid of them since paths are loaded based on present .traj files.

bruingineer:
allows paths without .traj files to be deleted without needing an app restart or to generate a .traj file

needs someone who knows the frontend JS/ts to OK this.

fixes #725


spacey-sooty (approved):
tested locally works

PR #732, opened 2024-09-17
Change to use 1d array in traj file samples

shueja:


src/document/path/ChoreoTrajStore.ts
```diff
@@ -47,49 +42,55 @@ export const ChoreoTrajStore = types
...
     }
   }))
   .actions((self) => ({
     deserialize(ser: Output) {
       self.waypoints = ser.waypoints;
+      self.splits = ser.splits;
       self.samples = ser.samples;
+      
       self.forcesAvailable = ser.forcesAvailable;
+      console.log(ser);
```
spacey-sooty, line 93: Needed?

oh-yes-0-fps (approved):
Lgtm, it looks like the split range is inclusive on both ends but double check if you haven't already.

shueja:
> Lgtm, it looks like the split range is inclusive on both ends but double check if you haven't already.

What do you mean by this?

spacey-sooty (approved):

PR #753, opened 2024-09-23
Cast solver exit condition to signed int

shueja:


trajoptlib/src/RustFFI.hpp
```diff
@@ -21,7 +21,7 @@ static void trycatch(Try&& func, Fail&& fail) noexcept try {
   fail(e.what());
 } catch (const sleipnir::SolverExitCondition& e) {
   // TODO: Use std::to_underlying() from C++23
-  fail(std::to_string(static_cast<uint8_t>(e)));
+  fail(std::to_string(static_cast<int8_t>(e)));
```
calcmogul, line 24: Also add `#include <type_traits>`.
```suggestion
  fail(std::to_string(static_cast<std::underlying_type_t<decltype(e)>>(e)));
```

oh-yes-0-fps (approved):
Happens to the best of us

calcmogul (changes requested):

trajoptlib/src/RustFFI.hpp
shueja: Can we do that now? I thought we didn't support c++23 fully yet

calcmogul: Reread the type more carefully.

calcmogul (approved):

trajoptlib/src/RustFFI.hpp
```diff
@@ -21,7 +22,8 @@ static void trycatch(Try&& func, Fail&& fail) noexcept try {
   fail(e.what());
 } catch (const sleipnir::SolverExitCondition& e) {
   // TODO: Use std::to_underlying() from C++23
-  fail(std::to_string(static_cast<uint8_t>(e)));
+  // numerical value of the enum, converted to string
+  fail(std::to_string(static_cast<std::underlying_type_t<decltype(e)>>(e)));
```
calcmogul, line 26:
```suggestion
  fail(std::to_string(
      static_cast<std::underlying_type_t<sleipnir::SolverExitCondition>>(e)));
```

calcmogul (changes requested):

calcmogul (approved):

PR #773, opened 2024-09-25
[choreo] Event marker refactor

shueja:


.vscode/settings.json
```diff
@@ -2,4 +2,100 @@
     "rust-analyzer.check.overrideCommand": [
         "cargo", "clippy", "--target-dir", "./target/clippy"
     ],
+    "files.associations": {
```
spacey-sooty, line 5: These should probably be in the user settings.json

shueja:
C++, Java and Python deserialization changes have been tested. Ready for review.

calcmogul (approved):
Does the event marker stuff have C++ tests?

calcmogul (changes requested):
Should add an assert to the Java test while you're at it. The commented out imports in ChoreoTests.java can also be removed.

calcmogul (approved):

PR #776, opened 2024-09-26
[choreolib] Downgrade to WPILib 2024

calcmogul:
WPILib 2025 beta won't be coming out until mid-October, and some teams want to test Choreo before vendors get around to making beta releases.


calcmogul:
Before Choreo is included in the WPILib beta installer, this PR should be reverted and a new release made.

PR #779, opened 2024-09-27
[chorelib] Fixup some things in choreolib java

oh-yes-0-fps:
- removed EXT types
- fixed an issue with null safety on module forces
- fixed the field name for timestamp in the samples
- fixed a guranteed recursion stack overflow in loading a traj from the cache


choreolib/src/main/java/choreo/autos/AutoTrajectory.java
```diff
@@ -135,6 +133,7 @@ private void logTrajectory(boolean starting) {
   }
 
   private void cmdInitialize() {
+    System.out.println("Starting trajectory " + name);
```
calcmogul, line 136: This debug print doesn't seem necessary.

choreolib/src/main/java/choreo/autos/AutoFactory.java
```diff
@@ -173,6 +173,9 @@ public <SampleType extends TrajectorySample<SampleType>> AutoFactory(
    * @see #voidLoop
    */
   public AutoLoop newLoop(String name) {
+    // causing a cache clear in simulation to allow a form of `hot-reloading`
+    // for trajectories in simulation
+    if (RobotBase.isSimulation()) clearCache();
```
calcmogul, line 178: This change should be ported to C++. We've also generally avoided single-line if statements in the rest of the codebase in favor of braces.

choreolib/src/main/java/choreo/autos/AutoTrajectory.java
```diff
@@ -73,7 +72,7 @@ <SampleType extends TrajectorySample<SampleType>> AutoTrajectory(
       ControlFunction<SampleType> controller,
       Consumer<ChassisSpeeds> outputChassisSpeeds,
       BooleanSupplier mirrorTrajectory,
-      Optional<TrajectoryLogger<SampleType>> trajectoryLogger,
+      Optional<TrajectoryLogger<SampleType>> trajLogger,
```
calcmogul, line 75: Revert this rename. Saving 6 letters isn't worth the loss in clarity.

calcmogul (changes requested):

choreolib/src/main/java/choreo/trajectory/DifferentialSample.java
```diff
@@ -17,7 +17,7 @@ public class DifferentialSample implements TrajectorySample<DifferentialSample>
       Choreo.getProjectFile().config.differentialTrackWidth.val;
 
   /** The timestamp of this sample, relative to the beginning of the trajectory. */
-  public final double timestamp;
+  public final double t;
```
spacey-sooty, line 20: Is there a way to work around having to name this t?

calcmogul: Is it due to serialization or something?

spacey-sooty: I think naming is due to serialisation

oh-yes-0-fps: All samples have a getTimestamp method aswell so i decided it was easier to change the name of the field in java opposed to everywhere else because there was already a clear way to get timestamp. I think we should reconsider trying to save space on the .traj files with shorter keys.

spacey-sooty (approved):
I think changing the name of fields is a seperate problem, since `getTimestamp` exists I'm fine with this going in.

PR #781, opened 2024-09-27
Removes stale indicator, path-wide detail settings, usesDefaultObstacles

shueja:
Path Config panel is temporarily disabled, but other path-wide configs could go there.
Stale indicator will be replaced with comparison against snapshot
usesDefaultObstacles was never used in logic.

### DIFF
--- a/src/components/sidebar/PathSelector.tsx
+++ b/src/components/sidebar/PathSelector.tsx
@@ -1,14 +1,7 @@
-import {
-  KeyboardArrowDown,
-  PriorityHigh,
-  Route,
-  Settings
-} from "@mui/icons-material";
+import { Route } from "@mui/icons-material";
 import DeleteIcon from "@mui/icons-material/Delete";
 import {
-  Checkbox,
   CircularProgress,
-  FormControlLabel,
   IconButton,
   TextField,
   Tooltip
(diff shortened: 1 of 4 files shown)


src/components/sidebar/PathSelector.tsx
```diff
@@ -191,7 +170,7 @@ class PathSelectorOption extends Component<OptionProps, OptionState> {
           }}
         ></TextField>
         <div>
-          <Tooltip disableInteractive title="Path Config">
+          {/* <Tooltip disableInteractive title="Path Config">
```
calcmogul, line 173: I'm not a fan of leaving dead code around.

src/components/sidebar/PathSelector.tsx
```diff
@@ -225,52 +204,13 @@ class PathSelectorOption extends Component<OptionProps, OptionState> {
           </Tooltip>
         </div>
         {/* Settings part */}
-        {this.state.settingsOpen && (
+        {/* {this.state.settingsOpen && (
```
spacey-sooty, line 207: Why keep this?

shueja: Because this is a preparation PR ahead of adding a setting for target dt that will go in that panel.

calcmogul (approved):

PR #796, opened 2024-10-01
Version to 2025.0.0-alpha1 and add version checks to choreolib

shueja:


choreolib/src/main/java/choreo/Choreo.java
```diff
@@ -147,6 +152,11 @@ static Trajectory<? extends TrajectorySample<?>> readTrajectoryString(
       String str, ProjectFile projectFile) {
     JsonObject wholeTrajectory = GSON.fromJson(str, JsonObject.class);
     String name = wholeTrajectory.get("name").getAsString();
+    String version = wholeTrajectory.get("version").getAsString();
+    if (!SPEC_VERSION.equals(version)) {
+      throw new RuntimeException(
+          name + ": Wrong trajectory version: " + version + ", expected " + SPEC_VERSION);
```
calcmogul, line 158:
```suggestion
          name + ": Wrong trajectory version: found " + version + ", expected " + SPEC_VERSION);
```

choreolib/src/main/java/choreo/Choreo.java
```diff
@@ -72,8 +73,12 @@ public static ProjectFile getProjectFile() {
       } else if (projectFiles.length > 1) {
         throw new RuntimeException("Found multiple project files in deploy directory");
       }
-      LAZY_PROJECT_FILE =
-          Optional.of(GSON.fromJson(new FileReader(projectFiles[0]), ProjectFile.class));
+      FileReader reader = new FileReader(projectFiles[0]);
+      JsonObject json = GSON.fromJson(reader, JsonObject.class);
+      if (!SPEC_VERSION.equals(json.get("version").getAsString())) {
+        throw new RuntimeException("Project file had wrong version, should be " + SPEC_VERSION);
```
calcmogul, line 79:
```suggestion
        throw new RuntimeException("Project file had wrong version. Found " + ? + "expected " + SPEC_VERSION);
```

calcmogul (changes requested):
choreolib/version.py will need to be modified to handle the custom tag format. WPILib uses tags like `v2025.0.0-alpha-1`. The python script will need to turn that into `2025.0.0a1` to comply with https://peps.python.org/pep-0440/#public-version-identifiers.

The alpha suffix is required so PyPI treats it as a pre-release.

choreolib/py/choreo/version.py
```diff
@@ -17,4 +17,6 @@ def get_version():
     if proc.returncode:
         return "0.0.0"
     else:
-        return re.search(r"v([0-9]+\.[0-9]+\.[0-9]+)", proc.stdout).group(1)
+        # Choreo uses tags like v2025.0.0-alpha-1. We turn that into 2025.0.0a1 
+        # to comply with https://peps.python.org/pep-0440/#public-version-identifiers.
+        return re.search(r"v([0-9]+\.[0-9]+\.[0-9]+)", proc.stdout).group(1).replace("-alpha", "a").replace("-beta", "b")
```
calcmogul: passing in `v2025.0.0-alpha-1` gives `2025.0.0`.

src-tauri/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "choreo"
-version = "2025.0.0-beta"
+version = "2025.0.0-alpha1"
```
calcmogul, line 3: Cargo uses version strings like `2025.0.0-alpha.1`.

choreolib/src/main/native/include/choreo/SpecVersion.h
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Choreo contributors
+
+#pragma once
+#include <string>
+
+namespace choreo {
+const std::string SPEC_VERSION = "v2025.0.0";
+}  // namespace choreo
```
calcmogul, line 8:
```suggestion
#pragma once

#include <string_view>

namespace choreo {

inline constexpr std::string_view SPEC_VERSION = "v2025.0.0";

}  // namespace choreo
```

shueja:
Cargo version number modification is not correct yet. I don't immediately know how to modify update_version.py to transform for the correct syntax.

choreolib/py/choreo/version.py
```diff
@@ -17,4 +17,18 @@ def get_version():
     if proc.returncode:
         return "0.0.0"
     else:
-        return re.search(r"v([0-9]+\.[0-9]+\.[0-9]+)", proc.stdout).group(1)
+        m = re.search(
+            r"^v([0-9]+\.[0-9]+\.[0-9]+)(-([0-9]+))?(-(.*))?",
```
calcmogul, line 21: .* is a greedy match, which matches too much (it matches all remaining characters, then gives back characters until the match succeeds).

The re.X argument is passed, so you can use whitespace to make the regex more legible.

Try some test strings in https://regex101.com/ to refine the regex.

choreolib/py/choreo/version.py
```diff
@@ -17,4 +17,25 @@ def get_version():
...
+
+        version = match.group(1)
+
+        # Choreo uses tags like v2025.0.0-alpha-1. We turn that into 2025.0.0a1
+        # to comply with https://peps.python.org/pep-0440/#public-version-identifiers.
+        # For dev: version number: <tag>.dev<# commits since tag>
+        if match.group(2):
+            version += f".dev{match.group(3)}"
+        elif match.group(4):
+            version += match.group(5).replace("alpha", "a").replace("beta", "b")
```
calcmogul, line 40:
```suggestion
            version += match.group(5).replace("alpha", "a").replace(
                "beta", "b"
            ) + match.group(7)
```

choreolib/src/main/native/include/choreo/Choreo.h
```diff
@@ -119,6 +125,10 @@ class Choreo {
   static std::optional<choreo::Trajectory<SampleType>> LoadTrajectoryString(
       std::string_view trajectoryJsonString) {
     wpi::json json = wpi::json::parse(trajectoryJsonString);
+    if (kSpecVersion != json["version"]) {
+      throw fmt::format("Trajectory has wrong version, should be {}",
```
calcmogul, line 129: You should also print the version it found instead for troubleshooting.

shueja:
Closes #793

choreolib/py/choreo/trajectory/__init__.py
```diff
@@ -122,6 +123,11 @@ def get_chassis_speeds(self) -> ChassisSpeeds:
 
         with open(chor, "r", encoding="utf-8") as project_file:
             data = json.load(project_file)
+        version = data["version"]
+        if version != SPEC_VERSION:
+            raise ValueError(
+                f"Project file was wrong version {version}. This ChoreoLib reads {SPEC_VERSION}"
```
calcmogul, line 129: The error messages between the three languages are inconsistent.

calcmogul (approved):

PR #801, opened 2024-10-01
Add a default active path

shueja:


src/document/DocumentManager.ts
```diff
@@ -610,6 +613,8 @@ export async function openProject(projectPath: OpenFilePayload) {
       throw "Internal error. Check console logs.";
     }
     doc.deserializeChor(project);
+    doc.pathlist.paths.clear();
+    console.log(trajectories);
```
calcmogul, line 617: Should this log call be here?

calcmogul (approved):

PR #819, opened 2024-10-03
[docs] Treat JavaDoc and Doxygen warnings as errors

calcmogul:


choreolib/build.gradle
```diff
@@ -47,6 +47,10 @@ dependencies {
     implementation 'edu.wpi.first.ntcore:ntcore-java:2024.+'
     implementation 'edu.wpi.first.ntcore:ntcore-jni:2024.+'
 
+    api "com.fasterxml.jackson.core:jackson-annotations:2.15.2"
```
oh-yes-0-fps, line 50: What do these `api` lines do?

calcmogul: They fix some javadoc warnings about missing Jackson stuff. WPILib uses Jackson in its user-facing API, but WPILib's build system doesn't handle transitive dependencies. Maven pom files normally declare those dependencies, but WPILib replaces them with blanks.

PR #820, opened 2024-10-03
[trajoptlib] Make GetIndex() usage consistent and fix GetIndex()

bruingineer:
GetIndex() previously had an increment on all calls to it. It also had inconsistent usage where uses assumed both 0 and 1 based indexing. This PR changes the function to be 0 indexed on waypoints and samples within segments.

the relevant code in swerve and differential drive generators have been updated. 

image


calcmogul (commented):
GenerateLinearInitialGuess still has a waypoint index for loop that starts at 1.

bruingineer:
> GenerateLinearInitialGuess() still has a waypoint index for loop that starts at 1.

That one is okay since it adds samples from wpt-1 to wpt. The call to GetIndex at the beginning of the function was adjusted.

spacey-sooty (approved):

PR #821, opened 2024-10-03
[choreo] [trajoptlib] Use more robust time estimates for intervals and initial dt decision variable

bruingineer:
blocked by #820

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,5 +2,4 @@
     "rust-analyzer.check.overrideCommand": [
         "cargo", "clippy", "--target-dir", "./target/clippy"
     ],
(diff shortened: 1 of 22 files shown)


shueja:
I can't fully test this until tomorrow, but looks logically correct.

spacey-sooty:
From my testing this has introduced a regression into the Differential generation (it can no longer generate a straight line path)

bruingineer:
I think the kinematics for differential generator need to get looked at.

spacey-sooty:
> I think the kinematics for differential generator need to get looked at.

The differential generator is fundamentally broken, but this PR caused a regression on top of that

spacey-sooty (approved):
Regressions appears to be gone

PR #856, opened 2024-10-10
Use marker name for triggers; change spec to prepare for timestamp-range markers.

shueja:
New Spec:
All markers in the `events` property of .traj root
```js
{
   name: "used for triggers",
   from: {
      offset: Expression
      targetTimestamp: number // offset.val + targetTimestamp = marker timestamp
      target: WaypointIDX // ignored by libraries
    },
    to: { // same as from, reserved for later},
    event: {/*pplib command with waitTime an Expression*/} | null //markers can have no gui-attached command.
}
```


spacey-sooty (approved):
Lgtm

PR #863, opened 2024-10-12
Re-enable opening of last project

shueja:
This was fully reimplemented, but disabled, probably as an artifact of a debugging session.


Lewis-Seiden (approved):
could be nice to have this be disableable in application settings

PR #881, opened 2024-10-19
[choreolib] Rename `AutoLoop` to `AutoRoutine`

oh-yes-0-fps:
Renamed `AutoLoop` to `AutoRoutine`
Fixed `auto-routines` doc examples after removal of `TriggerExt`


calcmogul:
Needs a C++ port.

oh-yes-0-fps:
Cpp would likely just store a `std::string_view` in the routine to do trajectory equality with most recent run in routine.

calcmogul:
std::string_view isn't an owning container, so it shouldn't be used as a member variable.

spacey-sooty:
Can you split AutoLoop -> AutoRoutine to a seperate PR?

choreolib/src/main/java/choreo/auto/AutoTrajectory.java
```diff
@@ -506,4 +506,13 @@ public Pose2d[] collectEventPoses(String eventName) {
...
     return poses;
   }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (obj instanceof AutoTrajectory) {
+      AutoTrajectory other = (AutoTrajectory) obj;
+      return this.name.equals(other.name);
+    }
+    return false;
```
spacey-sooty, line 516: This can be simplified I believe

spacey-sooty (changes requested):

spacey-sooty (approved):

PR #884, opened 2024-10-21
[choreolib] Fixed `done` logic

oh-yes-0-fps:
`done` is extensively document in its specific behavior and why it is the way it is.

The commit history looks weird here when un-squashed, this is built off of #881 with just reverting the commit in that where the change to `done` logic was reverted.
I'm sure there was a better way but idk how to do it

Requires #881 to be merged first


spacey-sooty (dismissed):
Rebase

spacey-sooty (changes requested):
I'd like a test to make sure this behaviour doesn't break

oh-yes-0-fps:
I ran it in sim in my robot project and there were no regressions. I will be running in actual matches in abt 8 hrs

spacey-sooty:
I mean like a unit test

choreolib/src/main/java/choreo/auto/AutoFactory.java
```diff
@@ -83,6 +83,11 @@ public void poll() {}
         @Override
         public void reset() {}
 
+        @Override
+        public boolean isMostRecentTrajectory(AutoTrajectory trajectory) {
+          return false;
+        }
```
spacey-sooty, line 89: Can you add a comment to explain this?

spacey-sooty (approved):

PR #895, opened 2024-10-29
Fix incorrect title of example in choreolib autoroutine docs

NoahStricker:
The title for the second example listed here https://sleipnirgroup.github.io/Choreo/choreolib/auto-routines/ seems to be mistakenly 

> Creating an auto routine with **composition** and a monolithic trajectory 

which is the same title as the last, (don't do this), example. I believe this was instead meant to say

>  Creating an auto routine with **triggers** and a monolithic trajectory


spacey-sooty (approved):
Thanks!

PR #902, opened 2024-10-31
[choreolib] removed `timeOffset` member from `AutoTrajectory`

oh-yes-0-fps:


calcmogul (changes requested):
Needs to be done for C++ as well.

PR #903, opened 2024-10-31
[choreolib] Cleaned up `AutoTrajectory` and `Trajectory` slightly

oh-yes-0-fps:
- changed `atTimeAndPlace` to `atTimeAndPose`
- removed unused `timeOffset` member from `AutoTrajectory`
- inlined some methods only used once
- added easier flipping to trajectory sample fetching
- added method to get the underlying trajectory of an `AutoTrajectory`


calcmogul (changes requested):
Needs C++ port.

calcmogul (approved):

PR #904, opened 2024-10-31
Replace no-args trigger factories(done(), active(), running(), etc.) with public properties

Daniel1464:
Resolves https://github.com/SleipnirGroup/Choreo/issues/897

Since these triggers are used so often, they can help significantly reduce clutter in the code.
Before:
```
path1.done().onTrue(...)
path2.active().onTrue(...)
```
Now:
```
path1.done.onTrue(...)
path2.active.onTrue(...)
```
This also reduces the number of allocations.

### DIFF
--- a/choreolib/src/main/java/choreo/auto/AutoFactory.java
+++ b/choreolib/src/main/java/choreo/auto/AutoFactory.java
@@ -82,11 +82,6 @@ public class AutoFactory {
 
         @Override
         public void reset() {}
-
-        @Override
-        public Trigger running() {
-          return new Trigger(loop, () -> false);
-        }
       };
 
   /** A class used to bind commands to events in all trajectories created by this factory. */
@@ -327,7 +322,7 @@ public class AutoFactory {
    */
   public AutoRoutine commandAsAutoRoutine(Command cmd) {
     AutoRoutine routine = newRoutine(cmd.getName());
-    routine.running().onTrue(cmd);
+    routine.running.onTrue(cmd);
     return routine;
   }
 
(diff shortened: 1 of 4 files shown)


Daniel1464:
Btw am working on c++ implementation soon; just kinda busy this week

spacey-sooty (changes requested):
This needs to update documentation as well

oh-yes-0-fps:
This should not be done, if you want to reduce allocations keep them stored as members internally.

Having access be a method is more flexible allowing for extra arguments, there really isn't a benefit to doing it like this besides removing the parent which isn't worth it for giving up coherency with the rest of the API.

spacey-sooty:
This restricts us in terms of future API changes so we won't do it but if you'd like to make these functions singletons that would be accepted

Daniel1464:
> Having access be a method is more flexible allowing for extra arguments, there really isn't a benefit to doing it like this besides removing the parent which isn't worth it for giving up coherency with the rest of the API.

I dont think this stops us from adding method overloads of done(), active(), etc if we need them to accept extra arguments. 

Having no-args methods be replaced by properties is also pretty consistent as well. In general, I think that if we don't need something to be a method it should not be a method, and it should be properly represented as what it is.

spacey-sooty:
> > Having access be a method is more flexible allowing for extra arguments, there really isn't a benefit to doing it like this besides removing the parent which isn't worth it for giving up coherency with the rest of the API.
> 
> I dont think this stops us from adding a done()/active() method that takes in arguments.
> 
> Having no-args methods be replaced by properties is also pretty consistent as well. In general, I think that if we don't need something to be a method it should not be a method, and it should be properly represented as what it is.

The issue with this is it makes it very hard to change the implementation of these Triggers without breaking the API. With a method this is reduced.

Daniel1464:
> The issue with this is it makes it very hard to change the implementation of these Triggers without breaking the API. With a method this is reduced.

ok honestly u guys prob have more foresight than me on this, so i'll trust your intuition.

PR #911, opened 2024-11-07
[choreolib] Make C++ compile again

spacey-sooty:
This also adds includes of Choreolib headers so we catch these in CI.

### DIFF
--- a/choreolib/src/test/native/cpp/main.cpp
+++ b/choreolib/src/test/native/cpp/main.cpp
@@ -7,7 +7,6 @@
 #include "choreo/auto/AutoChooser.h"
 #include "choreo/auto/AutoFactory.h"
 #include "choreo/auto/AutoRoutine.h"
-#include "choreo/auto/AutoRoutine.h"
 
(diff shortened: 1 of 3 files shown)


oh-yes-0-fps (approved):
Crazy that it didn't even compile b4

choreolib/src/test/native/cpp/main.cpp
```diff
@@ -2,6 +2,13 @@
 
 #include <gtest/gtest.h>
 
+#include "choreo/Choreo.h"
+#include "choreo/SpecVersion.h"
+#include "choreo/auto/AutoChooser.h"
+#include "choreo/auto/AutoFactory.h"
+#include "choreo/auto/AutoRoutine.h"
+#include "choreo/auto/AutoRoutine.h"
```
calcmogul, line 10: One of the includes was duplicated.

calcmogul (changes requested):

calcmogul (approved):

PR #915, opened 2024-11-11
Field overlay view layer for keep in/out zones

Lewis-Seiden:


src/components/field/svg/FieldOverlayRoot.tsx
```diff
@@ -242,6 +242,17 @@ class FieldOverlayRoot extends Component<Props, State> {
                 <FieldImage2024 />
               </>
             )}
+
+            {layers[ViewLayers.FieldOverlays] &&
+              doc.pathlist.activePath.params.constraints.map((c) => {
```
shueja, line 247: Can this be just the enabled constraints, and additionally filter out the one that's selected (if any)?

src/components/field/svg/constraintDisplay/KeepOutCircleOverlay.tsx
```diff
@@ -68,25 +70,32 @@ class KeepOutCircleOverlay extends Component<Props<"KeepOutCircle">, object> {
     const x = data.props.x.val;
     const y = data.props.y.val;
     const r = data.props.r.val;
+    console.log(doc.selectedSidebarItem);
+    const color =
+      uiState.layers[ViewLayers.Waypoints] &&
+      uiState.isNavbarWaypointSelected() &&
+      !this.props.selected
+        ? "rosybrown"
```
shueja, line 78: Not sure I like this color, but that's not a blocker.

src/document/UIData.tsx
```diff
@@ -148,6 +149,12 @@ export const ViewData = {
     name: "Focus",
     icon: <CropFree />,
     default: false
+  },
+  FieldOverlays: {
+    index: 6,
+    name: "FieldOverlays",
```
shueja, line 155: This naming is too vague and inconsistent with internal naming as well (everything is considered an Overlay).

shueja (changes requested):
Looks pretty good functionally. Just a few changes.

shueja:
Clicking a zone in the non-selected layer does not select it. If this is intentional, it's an unexpected break from existing UI norms within the app. If you're going to implement this, keep in mind that a click and drag on the non-selected zone will be hard to catch, since it will be on a different element than the one with the drag handler.

calcmogul:
Region or zone? TrajoptLib uses "keep-in region" and "keep-out region".

shueja:
It could be "region". As of current main, Choreo GUI doesn't have a word picked for this. Also worth noting that this new layer includes non-region constraints with display elements (currently just Point At) so we could swerve the question and call the view layer Constraints.

calcmogul:
"Constraints" sounds good to me.

oh-yes-0-fps:
@Lewis-Seiden what's up with this pr? was there anything that was non-functional or not up to par?

PR #926, opened 2024-11-15
[choreo] Document Updater

### LINKED ISSUE #570: Reimplement data upgrader in Rust for both types of 2025 files

oh-yes-0-fps:
Needs unit tests but has been tested on my robot project

This pr also adds `trackwidth` to .traj output field to show off the upgrader, its just a dummy value of 1.0 for now

Fixes #570.


calcmogul:
What are the steps for adding another schema version upgrade?

oh-yes-0-fps:
> What are the steps for adding another schema version upgrade?

Incrementing the constants for the schema version (its just an incrementing integer now for simplicity idk how you define semver for the schema stuff when any change is "breaking").
Add a new upgrader action in `upgraders.rs` for the changes you made
Update the choreolib code if its needed.

src-core/src/spec/trajectory.rs
```diff
@@ -399,6 +399,9 @@ pub struct Trajectory {
     /// This includes 0, but the index of the last sample is never in this list even if the split toggle is set
     /// for the last waypoint
     pub splits: Vec<usize>,
+    /// The trackwidth of the robot,
+    /// used in robot code for differential drive calculations
+    pub trackwidth: f64,
```
shueja, line 404: Either this should be optional or we have to define it for swerve

shueja (changes requested):
Architecturally this seems reasonable. I'll have to think about ramifications of changing save file versioning to an int. I also don't know if we should require that Choreolib changes matching spec changes must go in the same PR.

src-core/src/spec/trajectory.rs
oh-yes-0-fps: Yeah, requiring it for swerve isnt bad, the data is already in robot config

shueja: Is it? Which distance between modules is used?

spacey-sooty:
>  I also don't know if we should require that Choreolib changes matching spec changes must go in the same PR.

I think we risk creating issues/having Choreolib fall behind similar to how C++ does if it's not updated in the same PR.

shueja:
I agree. I'm just wondering what effect that will have on PRs getting carried through to mergeability.

calcmogul:
If we can't get it done in the PR itself, I'd be concerned about it ever getting done.

calcmogul:
Items to address:

- [ ] What is the "trackwidth" field when it's a swerve trajectory
- [ ] ChoreoLib C++/Java/Python changes to use an int for version number (and fail non-fatally if it's a string)
- [ ] Unit tests for the already implemented upgrader, and a clear place to put more

Item 3 is somewhat related to another issue about assembling a common set of test JSONs for all sorts of tests.

src-core/src/spec/trajectory.rs
oh-yes-0-fps: between front left and front right ig, we can document that behavior

shueja:
Trackwidth work moved to https://github.com/SleipnirGroup/Choreo/issues/969

calcmogul (approved):

PR #933, opened 2024-11-20
[choreolib] Upgrade to Gradle 8.11

calcmogul:


spacey-sooty (changes requested):
How many times did you run the wrapper task? Iirc there should br more changes than this

spacey-sooty (approved):

PR #946, opened 2024-11-23
[trajoptlib] Build and publish rust docs

### LINKED ISSUE #942: Build and publish rust docs for Trajoptlib

spacey-sooty:
Resolves https://github.com/SleipnirGroup/Choreo/issues/942

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -58,3 +58,8 @@ jobs:
 
       - run: mkdocs gh-deploy --dirty --force
         if: github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v')
+
+      - uses: actions/upload-artifact@v7
+        with:
+          name: Documentation
+          path: site
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -7,7 +7,7 @@ permissions:
 
 jobs:
   deploy:
-    runs-on: ubuntu-slim
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v7
 
@@ -36,10 +36,19 @@ jobs:
           sudo apt-get update -q
           sudo apt-get install -y doxygen
 
+      - name: Make GCC 14 the default toolchain (Linux)
+        if: runner.os == 'Linux'
+        run: |
+          sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-14 200
+          sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-14 200
+
       - uses: actions/setup-java@v5
         with:
           distribution: "temurin"
           java-version: 17
+
+      - name: Install Rust nightly
+        run: rustup toolchain install nightly --profile minimal --component clippy,rustfmt
 
       - name: Copy API docs
         run: ./make-docs.sh
--- a/docs/contributing/building-choreo.md
+++ b/docs/contributing/building-choreo.md
@@ -4,17 +4,17 @@
 
 ### Requirements for Windows
 
-- [node](https://nodejs.org/en/download/), [pnpm](https://pnpm.io/installation), [Rust](https://www.rust-lang.org/tools/install) ≥ 1.70.0, [CMake](https://cmake.org/download) ≥ 3.21, [Git](https://git-scm.com/)
+- [node](https://nodejs.org/en/download/), [pnpm](https://pnpm.io/installation), [Rust](https://www.rust-lang.org/tools/install) (nightly, installed automatically by `rustup` from `rust-toolchain.toml`), [CMake](https://cmake.org/download) ≥ 3.21, [Git](https://git-scm.com/)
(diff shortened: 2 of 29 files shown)


trajoptlib/include/trajopt/DifferentialTrajectoryGenerator.hpp
```diff
@@ -97,22 +97,22 @@ class TRAJOPT_DLLEXPORT DifferentialTrajectorySample {
   /// The heading.
   double heading = 0.0;
 
-  /// The left wheel velocity.
+  /// The left wheels velocity.
```
calcmogul, line 100: It was written that way because differential drives have only one wheel per side.

spacey-sooty: :woman_facepalming: I meant to add a an apostrophe

trajoptlib/src/error.rs
```diff
@@ -3,31 +3,46 @@ use thiserror::Error;
 // Error messages and codes from:
 // https://github.com/SleipnirGroup/Sleipnir/blob/v0.5.0/include/sleipnir/optimization/solver/exit_status.hpp
 
+/// Represents an error returned by TrajoptLib. Primarily exposes the Sleipnir [solver exit
+/// status](https://github.com/SleipnirGroup/Sleipnir/blob/v0.5.0/include/sleipnir/optimization/solver/exit_status.hpp).
```
calcmogul: Since the enum itself has docs now, the comment at the top of the file can be removed.

.github/workflows/docs.yml
```diff
@@ -36,6 +36,12 @@ jobs:
           sudo apt-get update -q
           sudo apt-get install -y doxygen
 
+      - name: Make GCC 14 the default toolchain (Linux)
+        if: runner.os == 'Linux'
+        run: |
+          sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-14 200
+          sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-14 200
```
calcmogul, line 43: Why is this needed?

trajoptlib/src/lib.rs
```diff
@@ -763,6 +978,16 @@ impl SwerveTrajectoryGenerator {
         );
     }
 
+    /// Keep the robot within a circle on the field over the continuum of state
+    /// between the waypoints at `from_index` and `to_index`. A point-point
+    /// maximum distance constraint is applied to the robot's origin and to
+    /// every corner of every bumper added with `set_bumpers`.
```
calcmogul: These docs would be a chore to change if the constraint's internal implementation ever does. Instead, focus on high-level, user-facing details. I'd modify the first sentence of the constraints to distinguish between the constraint applying to the robot's center or its bumpers (we don't have to mention it applying to both since the bumper region includes the center, from the user's point of view).

calcmogul (changes requested):

.github/workflows/docs.yml
spacey-sooty: `cargo doc` needs to build the rust code as it checks for compiler errors in example code listed in rustdoc.

spacey-sooty:
I've updated the rust toolchain to nightly that `cargo fmt` affects comments too.

docs/contributing/building-choreo.md
```diff
@@ -4,17 +4,17 @@
 
 ### Requirements for Windows
 
-- [node](https://nodejs.org/en/download/), [pnpm](https://pnpm.io/installation), [Rust](https://www.rust-lang.org/tools/install) ≥ 1.70.0, [CMake](https://cmake.org/download) ≥ 3.21, [Git](https://git-scm.com/)
+- [node](https://nodejs.org/en/download/), [pnpm](https://pnpm.io/installation), [Rust](https://www.rust-lang.org/tools/install) (nightly, installed automatically by `rustup` from `rust-toolchain.toml`), [CMake](https://cmake.org/download) ≥ 3.21, [Git](https://git-scm.com/)
```
calcmogul, line 7: Are there API stability implications to using nightly instead of stable? If it's a problem for deployments, we could use nightly just for docs linting.

trajoptlib/src/lib.rs
```diff
@@ -1068,6 +1431,18 @@ impl DifferentialTrajectoryGenerator {
         );
     }
 
+    /// Keep the robot between two lines parallel to a center line over
```
calcmogul:
```suggestion
    /// Keep the robot's center between two lines parallel to a center line over
```

calcmogul (changes requested):

docs/contributing/building-choreo.md
spacey-sooty: There are only concerns with specific nightly APIs, which I believe you need to enable even if you are running the nightly compiler, so it shouldn't be a major concern

calcmogul:
Could you modify the docs CI a bit to upload an archive of the built website as an artifact? That'll make docs changes easier to review in the future. Otherwise, looks good.

spacey-sooty:
Done, I've added --no-deps as well to keep the size of the uploaded artifact at a reasonable size

calcmogul (approved):

PR #947, opened 2024-11-24
Updated example docs

Daniel1464:
Replaced the orElseGet in the docs in favor of isEmpty() checks for the starting pose's Optional. 

Both of them work the same way, but an explicit isEmpty() check + returning a void routine makes it easier for rookie coders reading the documentation.

(Also corrects an invalid Commands.none() return in a method that returns an AutoRoutine)


docs/choreolib/auto-routines.md
```diff
@@ -89,26 +89,26 @@ public AutoRoutine fivePieceAutoTriggerSeg(AutoFactory factory) {
...
-                            return new Pose2d();
-                          }))
+          resetOdometry(initialPose.get())
               .andThen(
                   autoAimAndShoot(),
                   race(
                       intake(),
                       ampToC1.cmd(),
-                      aimFor(ampToC1.getFinalPose().orElseGet(Pose2d::new))))
+                      aimFor(ampToC1.getFinalPose().orElse(new Pose2d()))))
```
oh-yes-0-fps, line 111: Idk how this even got here but it shouldn't be here

Daniel1464: No yeah i replaced it on purpose; i think i should revert it tho

docs/choreolib/auto-routines.md
```diff
@@ -89,20 +89,20 @@ public AutoRoutine fivePieceAutoTriggerSeg(AutoFactory factory) {
...
       .onTrue(
-          resetOdometry(
-                  ampToC1
-                      .getInitialPose()
-                      .orElseGet(
-                          () -> {
-                            routine.kill();
-                            return new Pose2d();
-                          }))
+          resetOdometry(initialPose.get())
```
BR88C, line 105: If I'm not mistaken, this method is prone to error as it is only evaluating if the initial pose of the trajectory should be flipped when `fivePieceAutoTriggerSeg` is invoked. If the user is using `AutoChooser`, this would be when the auto is selected, at which point the robot may know the correct alliance. However assuming a worst case scenario where the user is utilizing the traditional `SendableChooser` or a custom solution, the segment would be initialized at startup. `resetOdometry()` should either be expected to accept a `Pose2d` supplier, or should be deferred. I understand this issue isn't the initial scope of the PR however I feel it is relevant.

docs/choreolib/auto-routines.md
```diff
@@ -168,20 +168,21 @@ public Command fivePieceAutoTriggerMono(AutoFactory factory) {
...
   routine.running()
       .onTrue(
-          resetOdometry(
-                  trajectory.getInitialPose()
-                      .orElseGet(
-                          () -> {
-                            routine.kill();
-                            return new Pose2d();
-                          }))
+          resetOdometry(initialPose.get())
```
BR88C, line 185: Ditto

docs/choreolib/auto-routines.md
```diff
@@ -227,15 +228,15 @@ public AutoRoutine fivePieceAutoCompositionSeg(AutoFactory factory) {
...
+  final Optional<Pose2d> initialPose = ampToC1.getInitialPose();
+  if (initialPose.isEmpty()) {
+      new Alert("Error: 5 piece auto has no starting pose", AlertType.kError)
+          .set(true);
+      return factory.voidRoutine();
   }
 
   Command ret = sequence(
-          resetOdometry(startinNoteOwnedose),
+          resetOdometry(initialPose.get()),
```
BR88C, line 239: Ditto

shueja (approved):

PR #951, opened 2024-11-25
[choreolib-java] removed sample array from trajectory

oh-yes-0-fps:
The user will handle this now, it could previously cause a runtime error.


spacey-sooty:
If we are doing this it should be documented how the user can do this.

oh-yes-0-fps:
They can figure out how to do `toArray`

spacey-sooty:
Samples method exists ig its fine

spacey-sooty (approved):

PR #953, opened 2024-11-25
[choreolib-java] made the trajectory cache more accesible

oh-yes-0-fps:
grr

### DIFF
--- a/choreolib/src/main/java/choreo/auto/AutoChooser.java
+++ b/choreolib/src/main/java/choreo/auto/AutoChooser.java
@@ -176,4 +176,14 @@ public class AutoChooser {
   public AutoRoutine getSelectedAutoRoutine() {
     return lastAutoRoutine;
   }
+
+
+  /**
+   * Get the {@link AutoFactory} used by this chooser.
+   * 
+   * @return The {@link AutoFactory} used by this chooser.
+   */
+  public AutoFactory factory() {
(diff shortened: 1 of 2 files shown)


choreolib/src/main/java/choreo/auto/AutoChooser.java
```diff
@@ -176,4 +176,13 @@ public void addAutoRoutine(String name, AutoRoutineGenerator generator) {
...
   }
+
+  /**
+   * Get the {@link AutoFactory} used by this chooser.
+   *
+   * @return The {@link AutoFactory} used by this chooser.
+   */
+  public AutoFactory factory() {
+    return factory;
+  }
```
spacey-sooty, line 187: Whys this needed?

oh-yes-0-fps: Incase you want to access the cache or smthn from the chooser, if a team wants to do that they would just assign the factory to a local var b4 passing it into the sender so this is just making it easier for teams. The factory is not a hidden or protected resource it's fine being publically accessible.

spacey-sooty: I was just curious as whenever I'd used it I'd had the factory accesible in the same scope as the chooser

spacey-sooty (approved):

Issue #944, opened 2024-11-23
Make doc links and vendordep JSON use new choreo.autos and lib.choreo.autos URLs


calcmogul:

shueja:
Closed by #955

PR #955, opened 2024-11-25
Switch to choreo.autos within docs in this repository

shueja:
Vendordep JSON should be changed with the next release.
Also the WPILib Vendordep Manager source should be updated.


shueja:
Closes #944

calcmogul (approved):

PR #957, opened 2024-11-27
[choreolib-java] Add support for alliance being unknown to AutoRoutine/AutoFactory

oh-yes-0-fps:


choreolib/src/main/java/choreo/auto/AutoRoutine.java
```diff
@@ -125,6 +134,10 @@ public void kill() {
    * @see #cmd(BooleanSupplier) A version of this method that takes a condition to finish the loop.
    */
   public Command cmd() {
+    if (!allianceKnown.getAsBoolean()) {
```
shueja, line 137: Should this run upon creating the command or upon scheduling it?

Daniel1464:
I'm not the biggest fan of Optional<Boolean>; at that point, you might as well use Optional<Alliance> instead.

PR #963, opened 2024-12-02
[choreolib] Add usage reporting

### LINKED ISSUE #962: Add usage reporting for trajectories and triggers
https://github.com/wpilibsuite/allwpilib/pull/7464:
```
kResourceType_ChoreoTrajectory = 119
kResourceType_ChoreoTrigger = 120
```

calcmogul:
Fixes #962.

### DIFF
--- a/.styleguide
+++ b/.styleguide
@@ -19,6 +19,7 @@ includeOtherLibs {
   ^frc/
   ^frc2/
   ^gtest/
+  ^hal/
   ^units/
   ^wpi/
 }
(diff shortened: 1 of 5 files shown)


shueja (approved):
1. Does Python have a way to do usage reporting in vendordeps?
2. It wasn't really clear to me from the name that ChoreoTrigger was for usage of the higher level API as a whole, but not much to do about that now.

calcmogul:
I checked RobotPy's API docs and couldn't find the HAL function required for usage reporting.

We could rename `ChoreoTrigger` if we want since it's our ID. Especially so since it's not in a WPILib release yet.

PR #970, opened 2024-12-04
[choreolib] Make AutoRoutine command end immediately if alliance optional is empty on schedule

shueja:
* AutoFactory now has a Supplier<Optional<Alliance>> which defaults to DriverStation.getAlliance()
* Teams now specify a BooleanSupplier useAllianceFlipping, instead of a supplier for mirroring.
* AutoRoutine.cmd() only runs if the alliance supplier returns a present optional (or flipping is disabled)


BR88C:
Is there a reason for `useAllianceFlipping` to be a supplier? What is the use case for changing flipping behavior during runtime?

choreolib/src/main/java/choreo/util/AllianceFlipUtil.java
```diff
@@ -226,4 +229,31 @@ public static Rotation3d flip(Rotation3d rotation) {
...
+   * Creates a Supplier<Optional<Pose2d>> based on a Supplier<Optional<Alliance>> and original Optional<Pose2d>
+   * @param pose The pose to flip
+   * @param alliance The current alliance
+   * @param doFlip Returns true if flipping based on the alliance should be done
+   * @return empty if the alliance is empty;
+   *  the original pose if the alliance is blue or doFlip is false;
+   *  the flipped pose if the alliance is red and doFlip is true
+   */
+  public static Supplier<Optional<Pose2d>> optionalFlipped(Optional<Pose2d> poseOpt, Supplier<Optional<Alliance>> allianceOpt, BooleanSupplier doFlip) {
+    return ()->allianceOpt.get().flatMap(
```
BR88C, line 243: Ditto, implementation should be `() -> doFlip.getAsBoolean() ? allianceOpt.get().flatMap(ally -> poseOpt.map(pose -> ally == Alliance.Red ? flip(pose) : pose)) : pose;`

shueja:
> Is there a reason for `useAllianceFlipping` to be a supplier? What is the use case for changing flipping behavior during runtime?

One use case is if you have some trajectories that should get mirrored and some that shouldn't (i.e. ones that are made for the red alliance), but that would be better served by some other API way of specifying per-trajectory. The original purpose of making it a supplier was for future flexibility, at FPS's suggestion.

docs/choreolib/auto-routines.md
```diff
@@ -196,14 +196,7 @@ public Command fivePieceAutoTriggerMono(AutoFactory factory) {
   // shoots the note when the shoot event marker is reached
   trajectory.atTime("shoot").onTrue(shootIfNoteOwned());
 
-  // aims the shooter when the aim event marker is reached,
-  // the aim command aims based on the next shoot event marker position
-  final AtomicInteger shootIndex = new AtomicInteger(0);
```
oh-yes-0-fps, line 201: Should maybe change from `onTrue(shootIfNoteOwned());` to like `onTrue(aim().andThen(shootIfNoteOwned()));`

spacey-sooty (approved):
Lgtm

BR88C (approved):
LGTM

shueja:
Don't merge yet, I'm adding tests for the alliance-based flipping

PR #971, opened 2024-12-04
[choreolib] Convert Trajectory API to Optionals

shueja:
Also refactor some sampleAt guard clauses into sampleInternal to make sampleInternal less reliant on preconditions.


choreolib/src/main/native/include/choreo/trajectory/Trajectory.h
```diff
@@ -95,14 +95,7 @@ class Trajectory {
...
-    std::optional<SampleType> state{};
-    if (samples.size() == 0) {
-      return {};
-    } else if (samples.size() == 1) {
-      return samples[0];
-    } else {
-      state = SampleInternal(timestamp);
-    }
+    std::optional<SampleType> state = SampleInternal(timestamp);
     if (state.has_value()) {
```
calcmogul, line 99:
```suggestion
    if (auto state = SampleInternal(timestamp)) {
```

choreolib/src/main/java/choreo/trajectory/Trajectory.java
```diff
@@ -169,26 +167,26 @@ public SampleType sampleAt(double timestamp, boolean mirrorForRedAlliance) {
    * @param mirrorForRedAlliance whether or not to return the pose as mirrored across the field
    * @return the initial pose of the trajectory.
    */
-  public Pose2d getInitialPose(boolean mirrorForRedAlliance) {
+  public Optional<Pose2d> getInitialPose(boolean mirrorForRedAlliance) {
     if (samples.isEmpty()) {
       return null;
```
calcmogul, line 172: This should return `Optional.empty()`.

calcmogul (changes requested):

calcmogul (approved):

PR #979, opened 2024-12-06
[choreolib] Add AutoRoutine.observe()

BR88C:
Adds `AutoRoutine.observe(BooleanSupplier condition)`, which returns a `Trigger` bound to the routine's event loop. Naming is flexible.

### DIFF
--- a/choreolib/src/main/java/choreo/auto/AutoRoutine.java
+++ b/choreolib/src/main/java/choreo/auto/AutoRoutine.java
@@ -108,6 +108,16 @@ public class AutoRoutine {
   }
 
   /**
+   * Creates a {@link Trigger} that is bound to the routine's event loop.
+   *
+   * @param condition The condition represented by the trigger.
+   * @return A {@link Trigger} that is {@code true} when the condition is satisfied.
+   */
+  public Trigger listen(BooleanSupplier condition) {
+    return new Trigger(loop, condition);
+  }
+
+  /**
    * Gets the poll count of the routine.
    *
    * @return The poll count of the routine.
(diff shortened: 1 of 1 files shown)


oh-yes-0-fps:
I'm still really not a big fan of the name, `.listen` makes me feel like I'm telling some action to listen for an event to happen and respond where this is really just sanitizing the scope of a `Trigger`. `.sanitize`, `.associate`, or maybe even like `.observe` if wanna stick with the implications that the `AutoRoutine` is doing something.

BR88C:
I like `observe()`, implementing now

choreolib/src/main/java/choreo/auto/AutoRoutine.java
```diff
@@ -107,6 +107,16 @@ public EventLoop loop() {
     return loop;
   }
 
+  /**
+   * Creates a {@link Trigger} that is bound to the routine's event loop.
+   *
+   * @param condition The condition represented by the trigger.
+   * @return A {@link Trigger} that is {@code true} when the condition is satisfied.
```
oh-yes-0-fps, line 114:
```suggestion
   * @return A {@link Trigger} that mirrors the state of the provided {@code condition}
```

choreolib/src/main/java/choreo/auto/AutoRoutine.java
```diff
@@ -107,6 +107,16 @@ public EventLoop loop() {
     return loop;
   }
 
+  /**
+   * Creates a {@link Trigger} that is bound to the routine's event loop.
```
oh-yes-0-fps, line 111:
```suggestion
   * Creates a {@link Trigger} that is bound to the routine's {@link EventLoop}.
```

choreolib/src/main/java/choreo/auto/AutoRoutine.java
oh-yes-0-fps: If you think this is worse just resolve the conversation

shueja (approved):

PR #981, opened 2024-12-06
Improve atPose clarity, routine-active safety, alliance-flipping safety

shueja:
The existing methods don't check rotation. TODO: add atPose that checks rotation

Also reorganize these to better handle alliance-dependent poses.

Makes `AutoTrajectory.allianceFlipped(Optional<Pose2d>)` and `allianceFlipped(Pose2d)' public.

### DIFF
--- a/choreolib/src/main/java/choreo/auto/AutoTrajectory.java
+++ b/choreolib/src/main/java/choreo/auto/AutoTrajectory.java
@@ -24,6 +24,7 @@ import java.util.ArrayList;
 import java.util.Optional;
 import java.util.function.BooleanSupplier;
 import java.util.function.Consumer;
+import java.util.function.Function;
 import java.util.function.Supplier;
 
 /**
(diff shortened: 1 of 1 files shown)


choreolib/src/main/java/choreo/auto/AutoTrajectory.java
```diff
@@ -529,16 +553,12 @@ public Trigger atPose(Optional<Pose2d> pose, double toleranceMeters) {
    * @param toleranceMeters The tolerance in meters.
    * @return A trigger that is true when the robot is within toleranceMeters of the given pose.
    */
-  public Trigger atPose(Pose2d pose, double toleranceMeters) {
-    return atPose(Optional.of(pose), toleranceMeters);
-  }
-
-  private Supplier<Optional<Pose2d>> optionalFlipped(Optional<Pose2d> pose) {
-    return AllianceFlipUtil.optionalFlipped(pose, alliance, useAllianceFlipping);
+  public Trigger atTranslationFlipped(Pose2d pose, double toleranceMeters) {
```
Daniel1464, line 556: I think u should rename "atTranslationFlipped" to "atTranslation" and "atTranslation" to "atAbsoluteTranslation";
most of the time, people want the pose to be flipped.

For it's namesake, an atTranslation(Translation2d) overload should exist too.

Daniel1464 (changes requested):

PR #982, opened 2024-12-07
Improve atPose clarity, routine-active safety, alliance-flipping safety

shueja:
Renamed atPose() triggers as atTranslation(), added new atPose() that requires rotation tolerance as well.
Fixed logic in those triggers that sampled an alliance-dependent pose too early
Made the triggers and(active()).

### DIFF
--- a/choreolib/src/main/java/choreo/auto/AutoTrajectory.java
+++ b/choreolib/src/main/java/choreo/auto/AutoTrajectory.java
@@ -460,22 +460,27 @@ public class AutoTrajectory {
     return trig;
   }
 
-  private Trigger atPose(Supplier<Optional<Pose2d>> pose, double toleranceMeters, double toleranceRadians) {
+  private Trigger atPose(
+      Supplier<Optional<Pose2d>> pose, double toleranceMeters, double toleranceRadians) {
     return new Trigger(
-      routine.loop(),
-      () -> {
-        Optional<Pose2d> checkedPoseOpt = pose.get();
-        return checkedPoseOpt
-            .map(
-                (checkedPose) -> {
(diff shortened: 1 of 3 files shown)


choreolib/src/main/java/choreo/auto/AutoTrajectory.java
```diff
@@ -499,22 +491,14 @@ public Trigger atTime(String eventName) {
...
-          Optional<Pose2d> checkedPoseOpt = checkedPoseOptSup.get();
-          return checkedPoseOpt
-              .map(
-                  (checkedPose) -> {
-                    Translation2d currentTrans = poseSupplier.get().getTranslation();
-                    return currentTrans.getDistance(checkedPose.getTranslation()) < toleranceMeters;
-                  })
-              .orElse(false);
-        });
+  public Trigger atPose(Optional<Pose2d> pose, double toleranceMeters, double toleranceRadians) {
```
oh-yes-0-fps, line 497: Why is this an optional? seems kinda leaky of the underlying implementation.

choreolib/src/main/java/choreo/util/AllianceFlipUtil.java
```diff
@@ -259,8 +259,48 @@ public static Supplier<Optional<Pose2d>> optionalFlipped(
    * @return empty if the alliance is empty; the original pose if the alliance is blue; the flipped
    *     pose if the alliance is red
    */
-  public static Supplier<Optional<Pose2d>> optionalFlipped(
+  public static Supplier<Optional<Pose2d>> optionalFlippedPose2d(
```
oh-yes-0-fps, line 262: All these optionals are confusing me

oh-yes-0-fps (approved):
looks good

PR #994, opened 2024-12-08
[choreolib, docs] ChoreoLib docs overhaul

BR88C:
The motivation here is to migrate to more of a "tutorial" style of documentation, as well as fix inconsistencies caused by recent API changes.

Primary changes:
- Article reorganization
- The trajectory controller is now documented in the "Getting Started" section, promoting discoverability, the Trajectory API now has its own article
- Differential drive control is now documented
- Trajectory API docs have been expanded to show proper use of a trajectory with a trajectory controller
- Major overhaul to the AutoFactory documentation
    - Majority of the documentation is now game agnostic, outside of branching autos
    - Improved documentation of individual features of the AutoFactory (instead of large code blocks)
    - Removal of large preamble in favor of generic subsystem names ("intakeSubsystem", "scoringSubsystem")
    - 2024 example now includes a graphic for pickup / scoring locations

Looking for feedback, some of the writing still feels too verbose.
 
To do:
- [x] Missing C++ code examples
- [x] Missing Python code examples
- [x] Odometry resetting example code (pending #995)
- [x] AutoChooser example code (pending #949)

PR #995, opened 2024-12-09
[choreolib] Add AutoTrajectory.resetOdometry()

BR88C:
Enforces that routines or command compositions will not execute if odometry is not seeded (trajectory cannot be loaded, is empty, or alliance is unknown), and reduces complexity for the user.

See the [this message](https://discord.com/channels/975739302933856277/1013180308117536788/1315321921222742067) in Discord for context.

Usage:
```java
AutoFactory autoFactory = Choreo.createAutoFactory(
  driveSubsystem::getPose,
  driveSubsystem::resetOdometry, // Consumer<Pose2d>
  driveSubsystem::followTrajectory
  // ...
);

// Using AutoRoutine
AutoRoutine routine = autoFactory.newRoutine("example");
AutoTrajectory myTrajectory = routine.trajectory("myTrajectory");
routine.running().onTrue(myTrajectory.resetOdometry().andThen(myTrajectory.cmd()));

// Using command composition
Commands.sequence(
  autoFactory.resetOdometry("myTrajectory"),
  autoFactory.trajectoryCommand("myTrajectory")
);
```

Still a little conflicted on the name. `seedOdometry()` feels a little more appropriate, however `resetOdometry()` is the nomenclature in the WPILib API so it's probably the better choice. Some interesting behavior to `AutoRoutine` could also be added to make resetting odometry automatic, though I'm also not sure how I feel about that. Feedback is welcome.

### DIFF
--- a/choreolib/src/main/java/choreo/Choreo.java
+++ b/choreolib/src/main/java/choreo/Choreo.java
@@ -329,6 +329,8 @@ public final class Choreo {
    * @param <SampleType> The type of samples in the trajectory.
    * @param poseSupplier A function that returns the current field-relative {@link Pose2d} of the
    *     robot.
+   * @param resetOdometry A function that receives a field-relative {@link Pose2d} to reset the
+   *     robot's odometry to.
    * @param controller A function that receives the current {@link SampleType} and controls the
    *     robot.
    * @param driveSubsystem The drive {@link Subsystem} to require for {@link AutoTrajectory} {@link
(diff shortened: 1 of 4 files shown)


BR88C:
Python build fail is seemingly unrelated.

Daniel1464:
I think it would be better to call this autoFactory.resetOdometryTo(path) and autoFactory.resetOdometryTo(pathName); as the auto factory does the path reset(not the trajectory itself).

BR88C:
> I think it would be better to call this autoFactory.resetPoseCmd(path) and autoFactory.resetPoseCmd(pathName); as the auto factory does the path reset(not the trajectory itself).

Not really sure what you're getting at here. Yes, the AutoFactory requires the `resetOdometry` consumer, however I don't think that should restrict ownership. Access via `AutoTrajectory` also reduces duplication if the user uses `AutoRoutine` (they won't have to specify the trajectory twice). Regardless, the logic should still exist in `AutoTrajectory` as it reduces complexity on the library side, if restricting access to only `AutoFactory` is desired then it could be made package private.

Appending "Cmd" or "Command" to the name makes sense, not sure which would be more appropriate as the abbreviation is inconsistent across the library (`AutoTrajectory.cmd()`, `AutoFactory.trajectoryCommand()`)

> Also, it might be good idea to have an autoFactory.resetPoseCmd(path, runnable) to let users handle an error scenario themselves

If the user wants custom behavior they should implement this themselves. Doubling the current set of overloads in `AutoFactory` seems unnecessary.

spacey-sooty (approved):
This makes sense to me, thanks! I assume docs will be updated in https://github.com/SleipnirGroup/Choreo/pull/994 ?

BR88C:
Yes that's the plan

PR #997, opened 2024-12-09
[choreolib] Replace DriverStation.reportWarning/reportError with alerts

Daniel1464:
Replaces DriverStation.reportWarning/reportError calls with wpilib's alerts api.

Fixes #996.

PR #999, opened 2024-12-10
[choreolib-java] move alert machinery to a util file

oh-yes-0-fps:
idk how this stuff gets approved


choreolib/src/main/java/choreo/Choreo.java
```diff
@@ -16,14 +16,15 @@
 import choreo.trajectory.SwerveSample;
 import choreo.trajectory.Trajectory;
 import choreo.trajectory.TrajectorySample;
+import choreo.util.AlertUtil;
+import choreo.util.AlertUtil.*;
```
calcmogul, line 20: Apparently we don't do wildcard imports.
https://github.com/SleipnirGroup/Choreo/pull/997/files/75da3b3e6b15babd09a34c823c95cd91113a5565#diff-429eb6d933fd118960d151d5273731da1a61957530710958ceb39fc48fd6ff86

Daniel1464:
Yeah its prob better to move these to a separate class. I think it might be better to call it "ChoreoInternal" though, since "AlertUtil" feels too close to something from wpilib.

Daniel1464:
> If it's for internal use only, make it package private.

You can't do that since choreo has submodules, and those submodules will not be able to access the package private class

Daniel1464:
Hey fps, btw could u rename the "Choreo" group in the AlertUtil.alert() and AlertUtil.multiAlert() methods to "ChoreoAlerts" instead? I realized the alerts are published under "SmartDashboard/Choreo" and not "SmartDashboard/Alerts/Choreo", which could be unclear to people what to do

oh-yes-0-fps:
> If it's for internal use only, make it package private.

package private to what? not all of our classes are in 1 package

spacey-sooty (approved):

Daniel1464 (approved):

PR #1003, opened 2024-12-11
[choreolib] Replace Choreo.createAutoFactory method with constructor

Daniel1464:
Choreo.createAutoFactory() didn't do anything that the constructors couldn't already do.

Note: I still kept in a no-args createAutoFactory() method to give a proper deprecation notice to people who were using the choreolib betas, just in case of confusion.

### DIFF
--- a/choreolib/src/main/java/choreo/Choreo.java
+++ b/choreolib/src/main/java/choreo/Choreo.java
@@ -5,7 +5,6 @@ package choreo;
 import static edu.wpi.first.util.ErrorMessages.requireNonNullParam;
 import static edu.wpi.first.wpilibj.Alert.AlertType.kError;
 
-import choreo.auto.AutoFactory;
 import choreo.trajectory.DifferentialSample;
 import choreo.trajectory.EventMarker;
 import choreo.trajectory.ProjectFile;
--- a/choreolib/src/main/java/choreo/Choreo.java
+++ b/choreolib/src/main/java/choreo/Choreo.java
@@ -324,15 +324,6 @@ public final class Choreo {
   }
 
   /**
-   * @Deprecated Use ```new AutoFactory()``` instead.
-   */
-  @Deprecated(forRemoval = true)
-  public static <SampleType extends TrajectorySample<SampleType>> AutoFactory createAutoFactory() {
-    throw new RuntimeException(
-        "Choreo.createAutoFactory is no longer supported; use new AutoFactory() instead.");
-  }
-
-  /**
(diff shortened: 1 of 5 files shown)


choreolib/src/main/java/choreo/Choreo.java
```diff
@@ -335,124 +324,12 @@ public void clear() {
...
-    return new AutoFactory(
-        requireNonNullParam(poseSupplier, "poseSupplier", "Choreo.createAutoFactory"),
-        requireNonNullParam(resetOdometry, "resetOdometry", "Choreo.createAutoFactory"),
-        requireNonNullParam(controller, "controller", "Choreo.createAutoFactory"),
-        requireNonNullParam(driveSubsystem, "driveSubsystem", "Choreo.createAutoFactory"),
-        requireNonNullParam(useAllianceFlipping, "useAllianceFlipping", "Choreo.createAutoFactory"),
-        requireNonNullParam(bindings, "bindings", "Choreo.createAutoFactory"),
-        Optional.of(trajectoryLogger),
-        requireNonNullParam(alliance, "alliance", "Choreo.createAutoFactory"));
+  @Deprecated(forRemoval = true)
```
calcmogul, line 329: If this is being deprecated, when are we removing it? Having a deprecation cycle for things in beta defeats the purpose of beta (being able to make API-breaking changes in response to feedback). I'd rather just delete it outright.

choreolib/src/main/java/choreo/Choreo.java
```diff
@@ -335,124 +324,12 @@ public void clear() {
...
-   *     be called every loop during the command.
-   * @param bindings Universal trajectory event bindings.
-   * @param trajectoryLogger A {@link TrajectoryLogger} to log {@link Trajectory} as they start and
-   *     finish.
-   * @param alliance A custom supplier of the current alliance to use instead of {@link
-   *     DriverStation#getAlliance}.
-   * @return An {@link AutoFactory} that can be used to create {@link AutoRoutine} and {@link
-   *     AutoTrajectory}.
-   * @see AutoChooser using this factory with AutoChooser to generate auto routines.
+   * @Deprecated Use ```new AutoFactory()``` instead.
```
calcmogul, line 327: javadoc requires all the fields to be here, so you can't just delete all of them. Also, javadoc uses `@deprecated` instead of `@Deprecated`.

calcmogul (changes requested):

calcmogul (approved):

PR #1005, opened 2024-12-12
Add a release yml for updating version numbers

shueja:
This updates all the version numbers and then commits to main and tags the commit made by the CI run. This will trigger the rest of our release process. The tag can still be slid later.


shueja:
Once merged, needs a PAT to be added to the repository so that Actions-made commits can trigger other Actions.

PR #1007, opened 2024-12-13
[choreo] Add float rounding in jsons

### LINKED ISSUE #1006: CLI and GUI generate files with slight differences
I was hoping to make a CI action to run the CLI generation and compare it against what was checked in our repo. We had a couple instances last year of people forgetting to hit the generate button, or generating with an older version of the GUI.

However, there are super small differences between what gets written when you hit the button in the gui, and run cli. WIthout having determinism between the two tools this CI checking idea is not going to work well. I have attached a diff of one of our files. 

This happens with beta6 while running with `--chor y2024/Crescendo/src/main/deploy/choreo/ChoreoAutos.chor --generate --all-trajectory`

[example.patch](https://github.com/user-attachments/files/18120569/example.patch)

oh-yes-0-fps:
closes #1006

Implemented into the json formatter


spacey-sooty:
Should update test json files

spacey-sooty (approved):
Worked for me logic makes sense

PR #1017, opened 2024-12-14
Rearrange the robot config panel.

shueja:
* Motor Calculator is gone
* Theoretical now always visible, shows torque and traction limits

image


spacey-sooty (approved):
This worked great for me and I see no obvious issues in the code

PR #1025, opened 2024-12-16
[choreolib-java] Change alliance flipping checks away from a supplier

oh-yes-0-fps:


choreolib/src/main/java/choreo/auto/AutoFactory.java
```diff
@@ -145,7 +144,7 @@ public <SampleType extends TrajectorySample<SampleType>> AutoFactory(
     this.resetOdometry = resetOdometry;
     this.controller = controller;
     this.driveSubsystem = driveSubsystem;
-    this.useAllianceFlipping = useAllianceFlipping;
+    this.useAllianceFlipping = () -> useAllianceFlipping;
```
Daniel1464, line 147: This should probably be changed to a boolean as well

oh-yes-0-fps: I don't want to have to change the underlying behavior yet cuz its a really pervasive change and im tired. I just need the public facing api to stabilize on what it will end up being.

choreolib/src/test/java/choreo/ChoreoTests.java
```diff
@@ -56,10 +54,10 @@ public class ChoreoTests {
...
 """;
 
-  private static final Gson GSON =
-      new GsonBuilder()
-          .registerTypeAdapter(EventMarker.class, new EventMarker.Deserializer())
-          .create();
+  // private static final Gson GSON =
+  //     new GsonBuilder()
+  //         .registerTypeAdapter(EventMarker.class, new EventMarker.Deserializer())
+  //         .create();
```
shueja, line 60: What's this commenting for?

shueja (changes requested):

choreolib/src/test/java/choreo/ChoreoTests.java
oh-yes-0-fps: It was unused. Idk why its unused now vs b4 tbh, just commenting it for now

shueja (approved):

PR #1026, opened 2024-12-16
Fix trajectory sample flipping and add tests

shueja:


choreolib/src/test/native/cpp/SampleFlippingTests.cpp
```diff
@@ -0,0 +1,175 @@
...
+                                   9_N,
+                                   10_N};
+
+    ASSERT_TRUE(sample.Flipped<2024>() == mirrored2024);
+    ASSERT_TRUE(sample.Flipped<2022>() == rotated2022);
+  } catch (std::exception& e) {
+    std::cerr << "Error: " << e.what() << std::endl;
+    FAIL();
+  }
+  SUCCEED();
```
calcmogul, line 174: I think this is redundant.

oh-yes-0-fps (approved):
java looks correct, probably should get feedback from tyler

calcmogul (approved):

PR #1030, opened 2024-12-16
Remove floating equality checks in ChoreoLib

shueja:


choreolib/src/main/java/choreo/trajectory/SwerveSample.java
```diff
@@ -320,17 +320,40 @@ public boolean equals(Object obj) {
...
+    && MathUtil.isNear(this.y       , other.y, 1E-9)
+    && MathUtil.isNear(this.heading , other.heading, 1E-9)
+    && MathUtil.isNear(this.vx      , other.vx, 1E-9)
+    && MathUtil.isNear(this.vy      , other.vy, 1E-9)
+    && MathUtil.isNear(this.omega   , other.omega, 1E-9)
+    && MathUtil.isNear(this.ax      , other.ax, 1E-9)
+    && MathUtil.isNear(this.ay      , other.ay, 1E-9)
+    && MathUtil.isNear(this.alpha      , other.alpha, 1E-9);
+    if (!equal) return false;
+    if (!fxBothNull) {
```
calcmogul, line 338: You could refactor this to use early exits instead to avoid nesting. Also, the array equals could be refactored into a utility function.

PR #1032, opened 2024-12-17
Replace Arrays.equals with primitive comparison for Java SwerveSample.equals

shueja:
Blocked on #1026 because after this change, flipping an array of 0.0 module force component should be equal to an array of 0.0, where in #1026 we have -0.0.

Arrays.equals uses Double.compare, which considers `-0.0 < 0.0`. We want our module forces to consider `-0.0 == 0.0`, thus they should use primitive `==`.


choreolib/src/main/java/choreo/util/ArrayUtil.java
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) Choreo contributors
+
+package choreo.util;
+
+import java.util.function.BiFunction;
+
+public class ArrayUtil {
+  /**
+   * Checks two <code>double</code> arrays for equality with the given function. This returns true
+   * if:
```
calcmogul, line 10: This needs an outer `<ul>` tag.

choreolib/src/main/java/choreo/util/ArrayUtil.java
```diff
@@ -0,0 +1,42 @@
...
+   * <li>Either both arrays are null, or
+   * <li>Neither is null, the arrays are the same length, and the given function returns true for
+   *     all same-index pairs of elements in the arrays.</>
```
calcmogul, line 13: I don't believe Javadoc needs a closing brace, but this wouldn't be a valid closing brace anyway.

calcmogul (dismissed):

oh-yes-0-fps (approved):
Tests pass, lgtm

PR #1033, opened 2024-12-17
Put the vendordep json in this repository

shueja:


.github/workflows/choreolib.yml
```diff
@@ -122,6 +122,16 @@ jobs:
         with:
           name: ${{ matrix.artifact-name }}
           path: choreolib/build/allOutputs
+  vendordep:
```
calcmogul, line 125: Newline before here to match the other job spacing.

calcmogul (changes requested):
The artifact name is a bit confusing.
Screenshot_20241217_192541

We could go with something like `ChoreoLib-json` so it's year-agnostic.

calcmogul (approved):

PR #1042, opened 2024-12-19
Automatic ChoreoLib Publishing

shueja:
Example of the outcome, before this workflow was restricted to tags: https://github.com/SleipnirGroup/ChoreoLib/pull/13
What the publishing workflow now skips if it's not a tag: https://github.com/SleipnirGroup/Choreo/actions/runs/12403982087/job/34628688962

### DIFF
--- a/.github/workflows/choreolib.yml
+++ b/.github/workflows/choreolib.yml
@@ -143,14 +143,24 @@ jobs:
         with:
           repository: wpilibsuite/build-tools
 
+      
+      - uses: actions/download-artifact@v4
+        with:
+          path: combiner/products/build/allOutputs
+      - name: Generate a token
+        # if: startsWith(github.ref, 'refs/tags/v')
+        id: generate-token
+        uses: actions/create-github-app-token@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/choreolib.yml
```diff
@@ -158,6 +184,24 @@ jobs:
       - name: Combine
         run: ./gradlew publish -Pthirdparty
         working-directory: combiner
+
+      - name: Transfer ChoreoLib Artifacts
+        if: startsWith(github.ref, 'refs/tags/v')
+        run: |
+          cp -r ~/releases/maven/development/choreo/* ChoreoLib/dep/choreo
+          pushd ./ChoreoLib
```
calcmogul, line 192: Is this `pushd` necessary? It's effectively just a `cd` command, and I was under the impression subsequent actions have independent shell contexts.

shueja: No. I'll remove it.

calcmogul (approved):

PR #1043, opened 2024-12-19
Add github token and persist-credentials:false to all actions/checkout

shueja:
https://yossarian.net/til/post/actions-checkout-can-leak-github-credentials/

Fortunately everything we could have leaked gets revoked at the end of the workflow, but I think this credential issue might be messing with our release version number updating workflow.


oh-yes-0-fps (approved):
idk what im looking at 👍

shueja:
> idk what im looking at 👍

The equivalent of turning off password autofill.

PR #1047, opened 2024-12-19
Normalize trackwidth spelling

calcmogul:


oh-yes-0-fps (changes requested):
This needs to have upgrader functionality added to support the rename in .chor

PR #1048, opened 2024-12-19
[Choreolib] cleaned up various parts of the non-user facing code

oh-yes-0-fps:


oh-yes-0-fps:
I goofed and the commit timeline is jank but the changes should be fine

choreolib/src/main/java/choreo/auto/AutoTrajectory.java
```diff
@@ -657,7 +635,7 @@ public Trigger atTranslation(String eventName, double toleranceMeters) {
 
     for (var event : trajectory.getEvents(eventName)) {
       // This could create a lot of objects, could be done a more efficient way
-      // with having it all be 1 trigger that just has a list of possess and checks each one each
+      // with having it all be 1 trigger that just has a list of posess and checks each one each
```
shueja, line 638:
```suggestion
      // with having it all be 1 trigger that just has a list of poses and checks each one each
```

choreolib/src/main/java/choreo/util/ChoreoAllianceFlipUtil.java
```diff
@@ -25,7 +24,11 @@
  * <p>This API still allows vendors and users to match case against the flipping variant as a way to
  * specially handle cases or throw errors if a variant is explicitly not supported.
  */
-public class AllianceFlipUtil {
+public class ChoreoAllianceFlipUtil {
+  /** An alias for {@code Supplier<Optional<Alliance>>} */
```
shueja, line 28: This doesn’t appear to be used.

shueja (changes requested):

choreolib/src/main/java/choreo/auto/AutoFactory.java
```diff
@@ -34,7 +34,7 @@
  */
 public class AutoFactory {
   static final AutoRoutine VOID_ROUTINE =
-      new AutoRoutine(null, "VOID-ROUTINE", new EventLoop()) {
+      new AutoRoutine(null, "VOID-ROUTINE", new EventLoop(), () -> true) {
```
bryceroethel, line 37: This is unsafe, as `trajectoryCmd()` will never fail even if the alliance is unknown (will default downstream to run a blue side auto). Not sure if there's an easy approach here as VOID_ROUTINE could no longer be static, see my comment in `AutoTrajectory` about a potential alternative.

Additionally, `VOID-ROUTINE` should be private. It isn't used elsewhere in the `choreo.auto` package, or for unit tests.

choreolib/src/main/java/choreo/auto/AutoRoutine.java
```diff
@@ -45,31 +48,22 @@ public class AutoRoutine {
...
    * @param factory The factory that created this loop
    * @param name The name of the loop
    * @param loop The inner {@link EventLoop}
+   * @param allianceKnownOrIgnored Returns true if the alliance is known or is irrelevant (i.e.
+   *     flipping is not being done).
+   * @see AutoFactory#newRoutine Creating a loop from a AutoFactory
    */
-  protected AutoRoutine(AutoFactory factory, String name, EventLoop loop) {
+  protected AutoRoutine(
+      AutoFactory factory, String name, EventLoop loop, BooleanSupplier allianceKnownOrIgnored) {
```
bryceroethel, line 62: `AutoRoutine` is initialized in two places: as `VOID-ROUTINE`, and in `newRoutine()`. In both cases, they effectively use `new EventLoop()`. `AutoRoutine.loop` should simply be initialized next to its declaration, and this constructor should be removed.

choreolib/src/main/java/choreo/auto/AutoTrajectory.java
```diff
@@ -127,7 +127,7 @@ <SampleType extends TrajectorySample<SampleType>> AutoTrajectory(
    * Returns true if alliance flipping is enabled and the alliance optional is present. Also returns
    * true if alliance flipping is disabled.
    */
-  private boolean allowSampling() {
+  private boolean allianceKnownOrIgnored() {
     return routine.allianceKnownOrIgnored.getAsBoolean();
```
bryceroethel, line 131: I'm not a fan of pulling protected variables from other classes, this should be injected in the constructor rather than pulling from `AutoRoutine`. Doing so would also resolve my comment about `VOID-ROUTINE` forcing `allianceKnownOrIgnored` to `true`. Consequently, protected fields of `AutoRoutine` should be made private.

Note that `routine.isActive` is used a lot in `AutoTrajectory`, which should be replaced with `routine.active().getAsBoolean()`.

choreolib/src/main/java/choreo/auto/AutoFactory.java
```diff
@@ -203,7 +203,7 @@ public AutoRoutine newRoutine(String name) {
    * @return True if the alliance is known or alliance flipping is disabled.
    */
   private boolean allianceKnownOrIgnored() {
```
bryceroethel, line 205: This is only used in `newRoutine()`, so it probably doesn't need its own method.

choreolib/src/main/java/choreo/util/ChoreoAllianceFlipUtil.java
oh-yes-0-fps: idek what that means

choreolib/src/main/java/choreo/auto/AutoTrajectory.java
```diff
@@ -223,15 +204,15 @@ public Command cmd() {
    *
    * @return A command that resets the robot's odometry.
    */
-  Command resetOdometry() {
+  public Command resetOdometry() {
```
bryceroethel, line 207: Docs need updating

choreolib/src/main/java/choreo/auto/AutoRoutine.java
```diff
@@ -28,62 +29,38 @@ public class AutoRoutine {
...
+  /** The alliance helper that is used to determine flipping logic */
+  private final AllianceContext allianceCtx;
 
   /** A boolean utilized in {@link #active()} to resolve trueness */
-  protected boolean isActive = false;
+  private boolean isActive = false;
 
   /** A boolean that is true when the loop is killed */
-  protected boolean isKilled = false;
+  boolean isKilled = false;
```
bryceroethel, line 47: Would it be useful to expose this to the user as a `killed()` method that returns a trigger?

choreolib/src/test/java/choreo/trajectory/SampleFlippingTest.java
```diff
@@ -4,14 +4,15 @@
...
+import choreo.util.ChoreoAllianceFlipUtil;
 import org.junit.jupiter.api.Test;
 
 public class SampleFlippingTest {
   private static final double FIELD_LENGTH_2022 = 16.5811;
   private static final double FIELD_LENGTH_2024 = 16.5811;
   private static final double FIELD_WIDTH_2022 = 8.19912;
-  private static final double FIELD_WIDTH_2024 = 8.19912;
+
+  //   private static final double FIELD_WIDTH_2024 = 8.19912;
```
bryceroethel, line 15: Is this meant to be left commented out or should it be removed?

bryceroethel (changes requested):

choreolib/src/test/java/choreo/trajectory/SampleFlippingTest.java
calcmogul: All commented-out code should be removed.

choreolib/src/main/java/choreo/auto/AutoRoutine.java
oh-yes-0-fps: what loop would that be polled on? kinda an anit-pattern

choreolib/src/main/java/choreo/auto/AutoFactory.java
```diff
@@ -141,34 +136,56 @@ public <SampleType extends TrajectorySample<SampleType>> AutoFactory(
...
+   * @param controller A function that receives the current {@link ST} and controls the robot.
    * @param driveSubsystem The drive {@link Subsystem} to require for {@link AutoTrajectory} {@link
    *     Command}s.
    * @param useAllianceFlipping If this returns true, when on the red alliance, the path will be
    *     mirrored to the opposite side, while keeping the same coordinate system origin.
    * @param bindings Universal trajectory event bindings.
    * @see AutoChooser using this factory with AutoChooser to generate auto routines.
    */
-  public <SampleType extends TrajectorySample<SampleType>> AutoFactory(
+  public <ST extends TrajectorySample<ST>> AutoFactory(
```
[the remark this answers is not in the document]
oh-yes-0-fps: It was done in other places by some other person idk who.
We had an issue with a user asking for help thinking `SampleType` was its own type, it's java convention that generic names are all caps and very short. Not really worth straying from that aslong as the generics are javadoc'ed

choreolib/src/test/java/choreo/trajectory/SampleFlippingTest.java
oh-yes-0-fps: I'll delete it

choreolib/src/main/java/choreo/auto/AutoRoutine.java
bryceroethel: Didn't think of that, ignore this then

bryceroethel (approved):
LGTM

choreolib/src/main/java/choreo/auto/AutoFactory.java
```diff
@@ -141,34 +136,56 @@ public <SampleType extends TrajectorySample<SampleType>> AutoFactory(
...
+            return new Trigger(this.loop(), () -> false);
+          }
+        };
   }
 
   /**
    * Create a factory that can be used to create {@link AutoRoutine} and {@link AutoTrajectory}.
    *
-   * @param <SampleType> The type of samples in the trajectory.
+   * @param <ST> The type of samples in the trajectory.
```
shueja, line 172:
```suggestion
   * @param <ST> The type of samples in the trajectory, either {@link DifferentialSample} or {@link SwerveSample}
```

choreolib/src/main/java/choreo/auto/AutoRoutine.java
```diff
@@ -18,8 +19,8 @@
 /**
  * An object that represents an autonomous routine.
  *
- * <p>This loop is used to handle autonomous trigger logic and schedule commands. This loop should
- * **not** be shared across multiple autonomous routines.
+ * <p>This routine is used to handle autonomous trigger logic and schedule commands. This routine
```
shueja, line 22:
```suggestion
 * <p>This object is used to handle autonomous trigger logic and schedule commands for a single autonomous routine. This object
```

choreolib/src/main/java/choreo/auto/AutoTrajectory.java
```diff
@@ -487,32 +466,10 @@ public Trigger atTime(String eventName) {
...
-  }
-
-  private Trigger atPose(Optional<Pose2d> pose, double toleranceMeters, double toleranceRadians) {
-    return atPose(
-        AllianceFlipUtil.optionalFlippedPose2d(pose, alliance, useAllianceFlipping),
-        toleranceMeters,
-        toleranceRadians);
+  private boolean withinTolerance(Rotation2d lhs, Rotation2d rhs, double toleranceRadians) {
+    double dot = lhs.getCos() * rhs.getCos() + lhs.getSin() * rhs.getSin();
+    double sinTolerance = Math.sin(toleranceRadians);
```
shueja, line 471: This doesn't seem right; `a dot b = ||a|| ||b|| cos theta`, not sin. In TrajoptLib we checked that the dot of the unit vectors was greater (closer to 1) than the cosine of the tolerance.

shueja (changes requested):

shueja (approved):

PR #1056, opened 2024-12-20
Remove schema versions from TypeScript

shueja:


src-core/src/spec/trajectory.rs
```diff
@@ -412,6 +412,7 @@ pub struct TrajectoryFile {
     /// Will always be in sync with the file name on disk.
     pub name: String,
     /// The version of the `.traj` file spec.
+    #[serde(default = "traj_schema_version")]
```
oh-yes-0-fps, line 415: I don't agree with this change, if the version isn't present its an invalid file because the upgrader doesn't know what actions to take.

src/document/2025/DocumentTypes.ts
```diff
@@ -137,7 +134,6 @@ export interface Output {
 
 export interface Trajectory {
   name: string;
-  version: typeof TRAJ_SCHEMA_VERSION;
```
oh-yes-0-fps, line 140: The frontend should still ensure the schema is complete using the type system as much as it can. Removing the version field can cause the `...` syntax for using values from one object to make another not carry over all the data the object may hold.

src/document/DocumentModel.tsx
```diff
@@ -89,7 +88,6 @@ export const DocumentStore = types
     serializeChor(): Project {
       return {
         name: self.name,
-        version: PROJECT_SCHEMA_VERSION,
```
oh-yes-0-fps, line 92: just change this value to a call from the backend or you can even lazily load the value at bootup and store it for the lifetime of the process.

oh-yes-0-fps (changes requested):

PR #1057, opened 2024-12-20
[choreolib] AutoChooser Fixes and Sim behavior changes

oh-yes-0-fps:
Makes changing alliance while disabled re-evaluate command creator
Makes going from disconnect->auto in sim quick load the auto then run it
Prevents a crash from composition going between enabled and disconnected
Fix an issue with `selected` in the sendables api being set when it shouldn't
Added `.instance` to work with elastic


oh-yes-0-fps:
The `.instance` isn't needed for glass but it is needed for elastic so this just adds a dummy value of 0 as per the [elastic source](https://github.com/Gold872/elastic-dashboard/blob/main/lib/widgets/nt_widgets/multi-topic/combo_box_chooser.dart) it's not used.

choreolib/src/main/java/choreo/auto/AutoChooser.java
```diff
@@ -43,17 +44,20 @@ public class AutoChooser implements Sendable {
   private final HashMap<String, Supplier<Command>> autoRoutines =
       new HashMap<>(Map.of(NONE_NAME, Commands::none));
 
-  // private final StringEntry selected, active;
-  // private final StringArrayEntry options;
+  private int allianceId = -1;
```
shueja, line 47: Why use a number here?

oh-yes-0-fps: For easier comparison

bryceroethel: What is wrong with using `.equals()`? As per `Optional`'s documentation:
> Indicates whether some other object is "equal to" this Optional. The other object is considered equal if:
    - it is also an Optional and;
    - both instances have no value present or;
    - the present values are "equal to" each other via equals().

shueja (approved):

PR #1059, opened 2024-12-20
Rename LocalProgressUpdate::DiffTrajectory to DifferentialTrajectory, fixing mismatch with TS

shueja:
This needs to match the camel-case transformation of [LocalProgressUpdate::DiffTrajectory](https://github.com/SleipnirGroup/Choreo/blob/37bd560cc8e035dd1a57030cf6d4d7b17159d5c9/src-core/src/generation/generate.rs#L32)


calcmogul (approved):
Fixes progress updates on my machine.

PR #1061, opened 2024-12-20
Fix Expression Input not updating: Revert #922 with better fix

shueja:
This bug happened when the input element was not recreated, but its expression was changed out. Most notably, this happens when going from one selected waypoint to another.

The fix is similar to 922, but keeps the old logic, except for recreating the input element whenever the expression serialization changes.


src/components/input/ExpressionInput.tsx
```diff
@@ -153,20 +165,37 @@ class Input extends Component<Props, State> {
...
+            this.setState({
+              editedValue: e.target.value
+            });
+            e.preventDefault();
           }}
-          onChange={() => this.updateValid()}
           onKeyDown={(e) => {
             if (e.key == "Enter") {
               this.inputElemRef.current?.blur();
+              // let newNumber = parseFloat(this.state.editedValue);
```
oh-yes-0-fps, line 187: These lines should be deleted

PR #1067, opened 2024-12-22
Remove numbers as valid input for dimensioned exprs

shueja:
This was causing footguns worse than its convenience.

### DIFF
--- a/src/document/ExpressionStore.tsx
+++ b/src/document/ExpressionStore.tsx
@@ -388,7 +388,6 @@ export const ExpressionStore = types
         tracing.error("evaluated to undefined or null");
         return undefined;
       }
-      console.log(newNumber);
       // numbers are only valid on dimensionless expressions.
       if (typeof newNumber === "number") {
         if (self.defaultUnit !== undefined) {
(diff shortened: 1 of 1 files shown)


src/document/ExpressionStore.tsx
```diff
@@ -376,54 +375,62 @@ export const ExpressionStore = types
...
           tracing.error("failed to evaluate: ", newNumber, "is infinite");
           return undefined;
         }
-        if (self.defaultUnit !== undefined) {
-          return addUnitToExpression(newNode, self.defaultUnit.toString());
-        }
+        // number is finite and on a dimensionless expression.
         return newNode;
       }
-      if (isNode(newNumber)) {
```
shueja, line 400: Artifact from when I misunderstood how variables got substituted in.

oh-yes-0-fps:
Does this mean that a number with no unit will not be accepted in fields that are unitly-dimensioned?

shueja:
> Does this mean that a number with no unit will not be accepted in fields that are unitly-dimensioned?

That is the intent.

oh-yes-0-fps (approved):

PR #1076, opened 2024-12-30
Make AutoBindings package private

spacey-sooty:


bryceroethel:
May be a good idea to update docs to reflect this change, i.e. modify the example `AutoFactory` instantiation to use the new constructor.

Additionally imho using `AutoFactory.bind()` is a better pattern than interfacing with an `AutoBindings` object directly, it may be worth discussing if the `AutoBindings` class should transition to becoming package private, and more of an implementation detail than a concern for the user.

spacey-sooty:
> May be a good idea to update docs to reflect this change, i.e. modify the example `AutoFactory` instantiation to use the new constructor.
> 
> Additionally imho using `AutoFactory.bind()` is a better pattern than interfacing with an `AutoBindings` object directly, it may be worth discussing if the `AutoBindings` class should transition to becoming package private, and more of an implementation detail than a concern for the user.

I like that as an approach.

spacey-sooty:
I think AutoBindings were missing from the AutoFactory ctor anyways in the docs...

docs/choreolib/auto-factory.md
```diff
@@ -229,14 +227,14 @@ public AutoRoutine branching2024Auto() {
 2. `AutoRoutine.anyActive()` can be used as a shorthand for checking if any trajectories in a set are active.
 3. Similar to `AutoRoutine.anyActive()`, `AutoRoutine.anyDone()` can be used as a shorthand for checking if any trajectories in a set are done.
 
-## AutoBindings
+## Auto Bindings
 
-The `AutoBindings` ([Java](/api/choreolib/java/choreo/auto/AutoFactory.AutoBindings.html)) class is used to bind event markers in trajectories made by the `AutoFactory` to commands. Commands added to `AutoBindings` exhibit the same behavior as those bound to `AutoTrajectory.atTime(String)`, except they are applied globally across all routines. This is useful if you have simpler actions that you want to trigger in any trajectory without much thought.
+Auto bindings are used to bind event markers in trajectories made by the `AutoFactory` to commands. Commands added to the `AutoFactory` using `bind` exhibit the same behavior as those bound to `AutoTrajectory.atTime(String)`, except they are applied globally across all routines. This is useful if you have simpler actions that you want to trigger in any trajectory without much thought.
```
shueja, line 232:
```suggestion
Auto bindings are used to bind event markers in trajectories made by the `AutoFactory` to commands. Commands added to the `AutoFactory` using `bind` exhibit the same behavior as those bound to `AutoTrajectory.atTime(String)`, except they are applied globally across all routines. This is useful if you have simpler actions that you want to trigger in any trajectory without much thought.

!!! warning
Even if a marker is bound individually in an `AutoTrajectory.atTime(String)` trigger, the global binding will still run, and cannot be disabled for a single marker.
```

shueja (changes requested):

shueja (approved):

PR #1080, opened 2024-12-31
Fix fmtJava pnpm script to run on powershell

shueja:
Tested on Windows Powershell and Windows Git Bash.

Previously `pnpm run fmtJava` called `bash` explicitly, but I found that Gradle handles the relative path to the `gradlew` if it's called like in this PR. Testing on other platforms is welcome.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   "type": "module",
   "scripts": {
     "fmtCpp": "wpiformat",
-    "fmtJava": "cd choreolib && bash ./gradlew spotlessApply",
+    "fmtJava": "cd choreolib && gradlew spotlessApply",
     "fmtJs": "prettier --write .",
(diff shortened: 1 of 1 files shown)


calcmogul:
This breaks in Bash.
```bash
[tav@myriad Choreo]$ pnpm run fmtJava

> choreo@2025.0.0-beta-9 fmtJava /home/tav/git/Choreo
> cd choreolib && gradlew spotlessApply

sh: line 1: gradlew: command not found
 ELIFECYCLE  Command failed.
 WARN   Local package.json exists, but node_modules missing, did you mean to install?
```

shueja:
That's what I thought, but then it worked in git bash. I imagine there isn't a fully cross-compatible way to make this work, so I'll close this.

PR #1082, opened 2024-12-31
Add a way to clear the path list that properly sets activePath to default

shueja:


src/document/DocumentModel.tsx
```diff
@@ -132,7 +132,6 @@ export const DocumentStore = types
       self.name = ser.name;
       self.variables.deserialize(ser.variables);
       self.robotConfig.deserialize(ser.config);
-      self.pathlist.paths.clear();
```
shueja, line 135: This was redundant since `openProject` deletes all the paths.

PR #1093, opened 2025-01-02
Display a different icon if the trajectory is changed since being generated

shueja:
The icon is clickable to trigger a regenerate.


calcmogul (approved):
Seems to do what's advertised on my machine. CI needs to be fixed though.

PR #1095, opened 2025-01-03
Refactor path gradients and make them properly support diff drive

calcmogul:
* Clean up implementations and comment their steps better
* Make linear velocity and linear acceleration gradients properly support differential drive
* Update API documentation


calcmogul:
The diff drive angular velocity gradient seems a bit sus. The diff drive project I've been testing basically looks like a rainbow.

calcmogul:
The rainbow was caused by the angular velocity going above the artificial max. I've increased it to 1 rotation per second.

PR #1099, opened 2025-01-03
Copy WPILib's CubicHermiteSpline as a differential initial guess

shueja:


calcmogul (approved):
Diff drive generator needs more testing in different scenarios, but I'm satisfied with the implementation.

PR #1116, opened 2025-01-06
Fix bad default path gradient

shueja:
Improves type safety on PathGradients


spacey-sooty (approved):
This fixed the behaviour I was seeing

PR #1117, opened 2025-01-06
[trajoptlib] Set dt upper bound

calcmogul:
Without it, the solver increases dt a lot, then reports infeasible on most paths.

The anti-tunneling constraint implicitly set an upper bound on dt, but setting an explicit upper bound is better.


spacey-sooty (approved):

shueja:
Not convinced this is the right maximum for everything but I haven't been able to break it yet.

PR #1118, opened 2025-01-06
Clear path list when opening a new project

shueja:
Missed this when adding `deleteAll()`


calcmogul (approved):
This fixed the issue for me.

PR #1120, opened 2025-01-06
Fix path export changes made in #1081

shueja:
Symptom: paths do not save when changes are made
Cause: Previously the PathListStore had an exporter function set by the DocumentManager on app initialization. It would pass this exporter to each added path. I missed removing this, so PathListStore was setting a no-op exporter to every path.
Fix/Prevention:
* Replaced `HolonomicPathStore.setExporter(function)` with `HolonomicPathStore.disableExport` which is only for the default path.
* Removed the path list's exporter function property entirely.

### DIFF
--- a/src/document/path/HolonomicPathStore.ts
+++ b/src/document/path/HolonomicPathStore.ts
@@ -197,15 +197,16 @@ export const HolonomicPathStore = types
         }
       );
     };
-    const setExporter = (exportFunction: (uuid: string) => void) => {
-      exporter = exportFunction;
+    /**Only to be used on the default path*/
+    const disableExport = () => {
+      exporter = () => {};
     };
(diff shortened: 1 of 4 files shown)


src/document/path/HolonomicPathStore.ts
```diff
@@ -197,15 +197,16 @@ export const HolonomicPathStore = types
         }
       );
     };
-    const setExporter = (exportFunction: (uuid: string) => void) => {
-      exporter = exportFunction;
+    /**Only to be used on the default path*/
```
shueja, line 200: image
The doc comment does show up when accessing the property, at least in VS Code.

calcmogul (approved):
Testing locally shows files save immediately upon changes, as expected.

Issue #1106, opened 2025-01-04
Switching to an ungenerated path mid-generation causes a UI crash


jlmcmchl:
See video in discord. https://discord.com/channels/975739302933856277/1192238434341896294/1325151778362228821

shueja:
I can't reproduce this particular bug, and the video didn't show if event markers were on any of the paths.

PR #1124, opened 2025-01-07
[UI] Improve undefined-safety when sampling a trajectory.

shueja:
This is the underlying exception-causer in #1106, but I don't know how the app got into the problem state in that issue.

### DIFF
--- a/src/components/field/svg/InterpolatedRobot.tsx
+++ b/src/components/field/svg/InterpolatedRobot.tsx
@@ -37,7 +37,7 @@ class InterpolatedRobot extends Component<Props, State> {
       return <></>;
     }
     const pose1 = sample(this.props.timestamp, trajectory);
-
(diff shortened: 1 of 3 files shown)


src/components/field/svg/InterpolatedRobot.tsx
```diff
@@ -37,7 +37,7 @@ class InterpolatedRobot extends Component<Props, State> {
       return <></>;
     }
     const pose1 = sample(this.props.timestamp, trajectory);
-
+    if (pose1 == undefined) return <></>;
```
Awesomeplayer165, line 40:
```suggestion
    if (pose1 === undefined) return <></>;
```

Awesomeplayer165 (changes requested):

Awesomeplayer165 (approved):

PR #1125, opened 2025-01-07
[ui] Make the sections of the variables panel clearer

shueja:
image
image


calcmogul:
I assume it's not supposed to look like this?
Screenshot_20250107_234518

calcmogul (approved):

PR #1131, opened 2025-01-07
Reduce pointless path renames when clicking in path selector

shueja:
Two issues:
1. when adding the icon that changes when generating and trajectory becoming stale, this started causing more re-renders of the whole path list row. I moved it to its own component that can re-render separately.
2. the path input now does not fire a rename unless the name has actually changed. Since people click on the path name and then click on the field, this was triggering renames that copied the file to the same path. These are now prevented.


oh-yes-0-fps (approved):
chbjbknm bcghjb bnvghujk
 bnvgyhjnb vghuj bgyuj bvgyuhjbnvgftyunbgf

PR #1133, opened 2025-01-08
Avoid using environment when deserializing commands before tree attachment

shueja:
Because we deserialize the command payload recursively, it's not in the main document tree at that time, so we can't use mobx environment injection to access the CommandStore constructor.

Prior bug: Try to load a project with an event marker which has at least one command payload (i.e. not "None). Loading fails with a browser console warning about "Cannot get environment for CommandStore".


oh-yes-0-fps (approved):
functions

PR #1149, opened 2025-01-15
[choreolib] Fix compilation error in Trajectory::GetTotalTime()

### LINKED ISSUE #1147: [choreolib] [cpp] Trajectory::GetTotalTime() doesn't unwrap optional, renders class virtually unusable
In `Trajectory.h`
```cpp
  /**
   * The total time the trajectory will take to follow
   *
   * @return The total time the trajectory will take to follow, if empty will
   * return 0 seconds.
   */
  units::second_t GetTotalTime() const {
    if (samples.size() == 0) {
      return 0_s;
    }
    return GetFinalSample().GetTimestamp();
  }
```

Code which uses GetTotalTime (or SampleAt, which calls into GetTotalTime) breaks due to the return statement in this method. GetFinalSample returns an optional so it should read

```cpp
return GetFinalSample().value().GetTimestamp();
```
The optional doesn't need to be checked since the method already verifies that the trajectory isn't empty.

I would submit a PR but I don't have the bandwidth right now to set up an entire testing environment just to add 8 characters to a file, so I apologize.

calcmogul:
Fixes #1147.

### DIFF
--- a/choreolib/src/main/native/include/choreo/trajectory/Trajectory.h
+++ b/choreolib/src/main/native/include/choreo/trajectory/Trajectory.h
@@ -10,6 +10,7 @@
 #include <utility>
 #include <vector>
 
+#include <units/time.h>
 #include <wpi/json_fwd.h>
 
 #include "choreo/trajectory/DifferentialSample.h"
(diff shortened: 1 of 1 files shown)


onthegrid007:
I just tried pulling the repo in vscode as a online vendor, how long untill these breaking changes get pushed downstream?
I also have an error for not Explicitly providing the Year parameter when calling Flipped(), I am not sure if i can manually fix these but if the changes were pushed i could just update the repo...

PR #1150, opened 2025-01-18
[choreo] Fix generation failure for zero angular velocity constraint between pose and translation waypoints

### LINKED ISSUE #1146: Generation fails for zero angular velocity constraint between pose and translation waypoints
> so I tried to make the robot not rotate between a pose and a translation waypoint
and this error popped up

image

bruingineer:
fixes #1146


spacey-sooty (approved):
Works great

PR #1151, opened 2025-01-18
[ui] Add a "Generate All Outdated" button

shueja:


spacey-sooty (approved):
works great, would be nice if the button was more obvious but idk where to put it thats better

PR #1152, opened 2025-01-18
[ui] Update staleness indicator when deserializing paths

shueja:
This is an explicit update because updating on changes to the serialization doesn't catch all paths when first opening the project.


shueja:
Testing:
on main, expect that when opening a project, all trajectories show as outdated, even if they're not actually outdated. This branch should fix that.

oh-yes-0-fps (approved):

PR #1153, opened 2025-01-18
Fix trajectoryCmd docs and leftover bug from #1110

shueja:
TBD if we want to have a separate command factory that doesn't go through AutoFactory at all.


shueja:
Obsolete by #1160

PR #1159, opened 2025-01-20
[choreolib] Added more triggers

oh-yes-0-fps:


choreolib/src/main/java/choreo/auto/AutoRoutine.java
```diff
@@ -188,16 +211,34 @@ public Trigger anyDone(AutoTrajectory trajectory, AutoTrajectory... trajectories
    * @param trajectory The first trajectory to watch.
    * @param trajectories The other trajectories to watch
    * @return a trigger that determines if any of the trajectories are finished
```
shueja, line 213:
```suggestion
   * @return a trigger that goes true for one cycle whenever any of the trajectories finishes, delayed by the given number of cycles.
```

shueja: I don't really know how to cleanly word this.

choreolib/src/main/java/choreo/auto/AutoTrajectory.java
```diff
@@ -199,6 +202,18 @@ public Command cmd() {
         .withName("Trajectory_" + name);
   }
 
+  /**
+   * Creates a command that will schedule <b>another</b> command that will follow the trajectory.
+   *
+   * <p>This can be useful when putting {@link AutoTrajectory} commands in sequences that require
+   * subsystems that the AutoTrajectory triggers require.
```
shueja, line 209:
```suggestion
   * subsystems also required by in AutoTrajectory-bound subsystems.
```

shueja (dismissed):
I don't think this should be merged until we root-cause #1161.

Daniel1464:
Maybe add a chain(Command, AutoTrajectory) overload? If that isnt clear enough you can call it thenRun instead of chain. 

Also you should document it too.

katzuv:
Is it possible to add `doneOverloads()` and `doneFor()` overloads that take time? And basically multiply the time by 50.

choreolib/src/main/java/choreo/auto/AutoRoutine.java
```diff
@@ -213,6 +255,21 @@ public Trigger anyActive(AutoTrajectory trajectory, AutoTrajectory... trajectori
     return trigger.and(this.active());
   }
 
+  /**
+   * Creates a trigger that returns true when any of the trajectories given are inactive.
+   *
+   * @param trajectory The first trajectory to watch.
+   * @param trajectories The other trajectories to watch
```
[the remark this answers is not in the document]
oh-yes-0-fps: To make sure it takes atleast 1 trajectory

spacey-sooty (approved):

PR #1163, opened 2025-01-20
[choreolib] fixes an issue with rerun trajectories not triggering done correctly

### LINKED ISSUE #1161: When an AutoRoutine is run a second time, trajectory .done() triggers do not fire.

oh-yes-0-fps:
closes #1161

The issue was the `inactive().and()` that the done trigger was wrapped in. This cause the triggers internal state to not be updated while the trajectory was running leading to the internal state not refreshing.

### DIFF
--- a/choreolib/src/main/java/choreo/auto/AutoFactory.java
+++ b/choreolib/src/main/java/choreo/auto/AutoFactory.java
@@ -150,7 +150,7 @@ public class AutoFactory {
 
           @Override
           public Trigger active() {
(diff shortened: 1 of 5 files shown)


shueja (commented):
Will test tonight on the setup I had issues with before.

shueja:
Found a way to reproduce the issue in sim and confirmed that this fixes it. I'm not sure if it breaks anything else though.

shueja (approved):

PR #1171, opened 2025-01-22
[trajoptlib] [choreolib] Fix bugs preventing 0 samples between waypoints

shueja:
This can happen with two identical adjacent waypoints, where the heuristic determines there is 0 time between them. This should be acceptable, because the two waypoints often do not have incompatible constraints. However, averaging segment time over the `0` samples in the segment led to NaNs and a "nonfinite cost or constraints" (actually nonfinite decision variable) solver error.

Accompanying ChoreoLib changes handle a split trajectory with no samples, which was mistakenly thought to become possible with this change. The code is left in for safety's sake, since the case was not already handled.

### DIFF
--- a/choreolib/src/main/java/choreo/trajectory/Trajectory.java
+++ b/choreolib/src/main/java/choreo/trajectory/Trajectory.java
@@ -239,13 +239,15 @@ public class Trajectory<SampleType extends TrajectorySample<SampleType>> {
    * @return a choreo trajectory that represents the split of the trajectory at the given index.
    */
   public Optional<Trajectory<SampleType>> getSplit(int splitIndex) {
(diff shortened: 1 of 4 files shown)


shueja:
I have tested that two identical pose waypoints in sequence generate on this PR and not on main, given that there is also another waypoint (so the trajectory has at least some samples). The failure feedback when the identical waypoints are the only waypoints in the path (thus a 1-sample path) is not useful, but that's a separate issue.

calcmogul (changes requested):
DifferentialTrajectoryGenerator.cpp needs the same changes.

calcmogul (approved):

PR #1180, opened 2025-01-23
[trajoptlib] Simplify callback code

calcmogul:


calcmogul:
I've confirmed on my Linux machine that progress animations and cancellation still work.

shueja (approved):
Still works for me on Windows as well.

PR #1186, opened 2025-01-29
[choreolib] Change auto checks to enabled

oh-yes-0-fps:


spacey-sooty (approved):

choreolib/src/main/java/choreo/auto/AutoRoutine.java
```diff
@@ -308,9 +308,9 @@ public Command cmd() {
    */
   public Command cmd(BooleanSupplier finishCondition) {
     return Commands.either(
-        Commands.run(this::poll)
+        Commands.startRun(this::poll, this::poll)
```
shueja, line 311: Why is this change related to the change in autonomous checks? It's a good change but needs its own PR and explanation of why.

shueja (changes requested):

PR #1189, opened 2025-02-02
[ci] Add pull request labeler

spacey-sooty:


.github/labeler.yml
```diff
@@ -0,0 +1,37 @@
...
+
+"component: choreolib-java":
+  - changed-files:
+      - any-glob-to-any-file:
+          ["choreolib/src/main/java", "choreolib/src/test/java"]
+
+"component: choreolib-cpp":
+  - changed-files:
+      - any-glob-to-any-file:
+          ["choreolib/src/main/native", "choreolib/src/test/java"]
```
calcmogul, line 13:
```suggestion
          ["choreolib/src/main/native", "choreolib/src/test/native"]
```

calcmogul (changes requested):
The labels should be alphabetically sorted.

allwpilib includes `**` in its globs. I'm not sure whether the glob patterns support partial matches.
https://github.com/wpilibsuite/allwpilib/blob/main/.github/labeler.yml

auscompgeek:
The readme seems to suggest the globs need to strictly match a changed filename. https://github.com/actions/labeler

.github/labeler.yml
```diff
@@ -0,0 +1,37 @@
...
+  - changed-files:
+      - any-glob-to-any-file: ["docs/**"]
+
+"component: ui":
+  - changed-files:
+      - any-glob-to-any-file: ["src/**"]
+
+"component: trajoptlib":
+  - changed-files:
+      - any-glob-to-any-file: ["trajoptlib/**"]
```
calcmogul, line 33: These need to be swapped for alphabetical sorting.

calcmogul (changes requested):

calcmogul (approved):

PR #1192, opened 2025-02-07
Make AutoTrajectory.cmd() not run if routine is inactive at initialize.

shueja:


shueja:
@oh-yes-0-fps as codeowner

oh-yes-0-fps (changes requested):
The `onlyIf` should go before the naming

PR #1193, opened 2025-02-07
fix: param order in AutoFactory

### LINKED ISSUE #1194: AutoFactory parameters are out of order

Sam948-byte:
closes #1194


Sam948-byte:
@shueja I think this might need to be rerun.  Looks like a random failure, given that my changes wouldn't have caused solely the MacOS build to fail.

PR #1196, opened 2025-02-07
[ChoreoLib] Fixed autochooser commands not ending with auto

oh-yes-0-fps:


spacey-sooty:
Do we want to handle this? or do we want to leave it to user code? both major ways of scheduling auto commands I can think of should handle this.

Issue #1197, opened 2025-02-08
Running CLI unlinks `pose.heading` variable


pjreiniger:
Tested with 2025.0.3 and previous versions.

I'm sure this is likely the same root cause as my other issue #1049, but this seems like a much higher priority problem. Up to this points my bugs have been about trivial floating point differences; either different sig figs down to 1e-6 or `-0 vs 0` but this once can actually cause real unexpected behavior.

Basically if you have a pose variable, x and y seem to get preserved when regenerating from the command line, but the `exp` for the heading field changes from `variable.heading` to the raw radians. This happens in the `params.waypoints` dict. The `snapshot.waypoints` is fine since it doesn't contain the expressions, only the values.

 Diff of pre-cli vs post-cli
![Image](https://github.com/user-attachments/assets/6b927eea-dc5e-4923-8fcb-f4fe46489a58)

Picture of path in the gui after cli
![Image](https://github.com/user-attachments/assets/003c7be5-845d-4079-84c8-493f9f741adb)

Between our several testing and tuning paths, and our actual auto paths we have over 70 paths. We made an incorrect assumption on whether our scoring mechanism would be on the front or the back of the robot, and it would be a whole lot easier to change our 14 variables heading and run the CLI was usable.

shueja:
Should be fixed by #1393. Must test and determine if further work in another PR is required.

Issue #1049, opened 2024-12-19
CLI and GUI still produce different results in beta8


pjreiniger:
Same idea as #1006.

It appears in beta8 that the the GUI is writing out `.traj` files with `"version": 1`, while the cli writes out `"version": 0`. There is also a problem with one writing out `0` and the other writing out `-0` in certain situations.

Attached diff from generating with the gui as the base, then running the cli on top
[beta8_example.patch](https://github.com/user-attachments/files/18202748/beta8_example.patch)

calcmogul:
Looks like we forgot to update the schema version somewhere.

We could just normalize `-0` to `0`.

pjreiniger:
The GUI writes out `version 0`*

This means that you can not deploy / simulate after opening the gui

shueja:
Fixed in beta 9

shueja:
> We could just normalize `-0` to `0` since they mean the same thing in floating point.

In Rust, `(-0.2f64).round() == -0.0`, so an extra normalization step is a valid thing to add.

shueja:
> Attached diff from generating with the gui as the base, then running the cli on top

Can you please clarify which platform the GUI and CLI are running on? Your setup from the other issue sounds like the GUI generates on a personal machine (windows, likely), and the CLI generates in a GitHub runner using a different platform.

We expect slightly different outcomes from the generation on different platforms (thus the rounding/normalization of differences below a given tolerance), but between the CLI and GUI on the same platform, they should be running the same code path and getting the same results.

pjreiniger:
At the moment I do this on my Windows 11 machine:
- Download the latest windows `setup.exe` (just tried beta 9), and install it.
- Run the gui, and regenerate every path
- Make a git commit
- Run the CLI
-Observe differences. As mentioned above, the big one now is rounding negative 0.

I know if that doesn't work, there is no hope for it to work in CI.

I would *think* that on different architectures should be able to produce identical values down to millimeter precision. It would be very annoying if my Mac kids cannot run choreo on their computer and have it produce deterministic results compared to our team owned driver station computers.

If needed I can easily run the CI action on windows once that starts to work instead of linux, but that was my plan.

pjreiniger:
I tried the last master build. Almost there

[still_bad.patch](https://github.com/user-attachments/files/18217822/still_bad.patch)


[...]

PR #1201, opened 2025-02-16
Fixup CLI modifying trajectory parameters

pjreiniger:
Fixes: #1197
Fixes: #1049

I tried to take the relevant parts from the [mega PR](https://github.com/SleipnirGroup/Choreo/pull/1070/files?diff=split&w=1#diff-0c1099e790aac17257467ee24a431097c49275ce79a8871a506f73cd463e1737R72-R88) that said it should fix the CLI problems.

I am not super well versed in Rust, so there might be a more rust-ic way to do the Result thing.

I tested this by running the `2025.0.3` gui on our paths last year and this year, making a commit, and then running `choreo-cli.exe --chor <file path> --generate --all-trajectory` and verifying there were no changes. Our paths aren't complicated, basically just stop points and max velocity constraints, but it works for those and doesn't break any tests.

If something does break from this, I would say [If you like it then you should have put a test on it](https://abseil.io/resources/swe-book/html/ch11.html#the_beyonceacutesemicolon_rule)

### DIFF
--- a/src-core/src/generation/generate.rs
+++ b/src-core/src/generation/generate.rs
@@ -115,6 +115,9 @@ pub fn generate(
     mut trajectory_file: TrajectoryFile,
     handle: i64,
 ) -> ChoreoResult<TrajectoryFile> {
+    let mut original = trajectory_file.clone();
+
+    // these two functions can make changes to the trajectory file
     set_initial_guess(&mut trajectory_file);
     adjust_headings(&mut trajectory_file)?;
 
@@ -125,5 +128,11 @@ pub fn generate(
     gen.add_omni_transformer::<ConstraintSetter>();
     gen.add_omni_transformer::<CallbackSetter>();
 
-    gen.generate()
+    let new_traj = gen.generate();
+    if new_traj.is_ok() {
+        original.trajectory = new_traj?.trajectory;
+        Ok(original)
+    } else {
+        new_traj
+    }
 }


src-core/src/generation/generate.rs
```diff
@@ -115,6 +115,9 @@ pub fn generate(
     mut trajectory_file: TrajectoryFile,
     handle: i64,
 ) -> ChoreoResult<TrajectoryFile> {
+    let mut original = trajectory_file.clone();
```
calcmogul, line 118:
```suggestion
    let mut original_file = trajectory_file.clone();
```

src-core/src/generation/generate.rs
```diff
@@ -125,5 +128,11 @@ pub fn generate(
...
     gen.add_omni_transformer::<CallbackSetter>();
 
-    gen.generate()
+    let new_traj = gen.generate();
+    if new_traj.is_ok() {
+        original.trajectory = new_traj?.trajectory;
+        Ok(original)
+    } else {
+        new_traj
+    }
```
calcmogul, line 137: This should be equivalent, but more Rust-y.
```suggestion
    match gen.generate() {
        Err(e) => Err(e),
        Ok(new_trajectory_file) => {
            original_file.trajectory = new_trajectory_file.trajectory;
            Ok(original_file)
        }
    }
```
Why does returning the original file instead of the new file here fix the problem?

calcmogul (changes requested):

pjreiniger:
After testing with some new paths we tried to generate, I've completely changed the strategy for this.

After some digging, the solution proposed in the mega PR and partially copied here is not enough to work for a trajectory that has just waypoints (generate has never been called in the GUI).

Unlike previously stated, the control flow for CLI and GUI generation is very different. When generating from the GUI, only partial results from the `generate` function are used, and then additional steps like stamping the snapshot and guessing control intervals are calculated. The CLI must be updated to follow this flow as well.

This seems very gross. One would think that calling `generate` would do all of the updates to the `TrajectoryFile` struct, and the UI would just show these updates instead of doing extra steps. However, as this is in the middle of the season and I think one of the maintainers should make such sweeping changes anyway, I went with the minimally invasive approach and updated the CLI tool to mimic the extra steps the gui does. Since the CLI is pretty hopelessly useless at the moment, I don't think this will break anyone's workflow, unlike making a change to the primary control flow.

This was tested by running a [python script](https://github.com/GirlsOfSteelRobotics/GirlsOfSteelFRC/blob/main/y2025/Reefscape/generate_choreo_mini_paths.py) that creates a `.traj` with only waypoints and constraints in it. I then verified that the results from hitting "generate all" in the GUI matched the results of running the CLI on all the trajectories.


Once again, I will say that I'm probably doing some very un-rust like things and that should be looked at.

shueja:
Close as addressed in #1394

PR #1213, opened 2025-03-06
No-op, not panic, if project was written before notifier wakeup

shueja:


shueja:
This is intended to fix issues with project files not saving.

calcmogul:
OBE by #1221.

PR #1223, opened 2025-04-07
[trajoptlib] Upgrade Sleipnir for faster solves

calcmogul:
Merge this after #1224.


calcmogul:
Merge this after #489 and #1224.

PR #1224, opened 2025-04-07
[trajoptlib] Use C++23

calcmogul:
Merge this after #489.


calcmogul:
Merge this after #489.

Issue #1202, opened 2025-02-22
Frontend error


calcmogul:
![Image](https://github.com/user-attachments/assets/4ede816e-f8ae-4fd2-bb88-4017f9c43af1)
[project.zip](https://github.com/user-attachments/files/18917479/project.zip)

calcmogul:
This was caused by "Back Mod X" not being negative. The app should enforce this.

PR #1242, opened 2025-04-24
Error to user when back mod x is positive

spacey-sooty:
Resolves https://github.com/SleipnirGroup/Choreo/issues/1202
image

### DIFF
--- a/src/components/input/ExpressionInput.tsx
+++ b/src/components/input/ExpressionInput.tsx
@@ -21,6 +21,8 @@ export type ExpressionInputProps = {
   titleTooltip?: string;
   /** Maximum width of the number input, in monospace characters */
   maxWidthCharacters?: number;
+  /** A function to verify the validity of an input. */
+  valid?: ((thing: IExpressionStore) => boolean);
 };
 
 type State = {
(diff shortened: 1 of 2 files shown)


src/components/input/ExpressionInput.tsx
```diff
@@ -21,6 +21,8 @@ type Props = {
   titleTooltip?: string;
   /** Maximum width of the number input, in monospace characters */
   maxWidthCharacters?: number;
+  /** A function to verify the validity of an input. */
+  valid: ((thing: IExpressionStore) => boolean) | undefined;
```
shueja, line 25:
```suggestion
  valid?: ((thing: IExpressionStore) => boolean);
```
As far as TS is concerned, you want `?` for something that doesn't need to be written, but `| undefined` if it must be written but could be `valid={undefined}` (kinda pointless here but more generally useful.

This is the issue giving CI errors.

shueja (changes requested):
Implementation looks to be in the right direction.

This system should be used for all the must-be-positive config values like mass, MoI, etc. 

Also this doesn't prevent a parseable but positive backModX from being submitted, does it? You would have to put the `this.props.valid` check in `this.getValid()` for that. At some point it makes more sense to have custom validity checks in the ExpressionStore itself, not the input element.

src/components/input/ExpressionInput.tsx
spacey-sooty: I was unclear about that, thanks for clarifying

PR #1249, opened 2025-04-27
Upgrade mobx-react to 9.2.0 with necessary architectural changes.

shueja:


shueja:
App doesn't white-screen on boot or use of waypoints, constraints, or markers, but I haven't thoroughly looked for other possibly-bugged uses of this.props.

calcmogul (approved):

PR #1255, opened 2025-05-16
Add logic to backend postprocess to update event marker timestamps

shueja:
Also comment nearby logic.


shueja:
Tested with the `src-core` integration tests which update the marker timestamps in the `test-jsons` test files with this change, and don't update them without this change.

calcmogul (approved):

PR #1257, opened 2025-05-16
Rework generation dataflow to avoid mutating parameters

shueja:
Previously:
* Mutate parameters to adjust headings (by changing the original heading expression) and identify initial guess points (by changing a nonserialized helper field on Waypoint)
* Take a snapshot of the mutated parameters and use that to set up the SwerveTrajectoryGenerator. 
* During TrajectoryGenerator setup, guess the control interval counts and handle them separately from the snapshot
* After generation, guess the control interval counts *again* and update both the snapshot and the params section with new counts.

This results in bugs where the generator wipes away the expressions in waypoint headings. This was identified as a CLI bug, but it is a core bug that's masked by the way the frontend handles generation results.

New behavior:
* Take a snapshot immediately of the params, never change the params again.
* Mutate clones of the snapshot to adjust headings, identify initial guess points, and store the control interval guess in the `snapshot.waypoints[n].intervals` fields. 
* Set up the TrajectoryGenerator with the modified snapshot.
* After generation, do not update the params section of the trajectory with the updated control interval counts. The snapshot section is the new source of truth for how waypoints correspond to the samples array.

Also this PR has a new docs page explaining what parts of the TrajectoryFile change upon generation.

### DIFF
--- a/docs/contributing/generation-invariants.md
+++ b/docs/contributing/generation-invariants.md
@@ -1,0 +1,30 @@
+# Generation Invariants
+
+> This is development documentation describing the dataflow as of 5/16/2025. It is not accurate for released versions.
+## Input
+`pre.traj` and `pre.chor`
+* presumed to already be schema-updated to current
+
+## Output
+`post.traj` as serialized  (file write or other response)
+
+`pre.chor` is unchanged
+    * Schema updates might cause the file contents to be replaced with the updated version
+
+
+
+## After Successful Trajectory Generation
+* `name` is unchanged
+* `version` is unchanged
+* `params` is unchanged
+* `snapshot` corresponds to the `params` field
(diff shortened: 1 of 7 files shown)


shueja:
I tested with the `src-core` integration tests that:
* on main, a `param.waypoint.heading` of {"0 deg", 0.0} was output as {"0 rad", 0.0}, which matches the buggy behavior this PR tries to fix.
* on this branch, the same 0 deg parameter is output unchanged, as expected.
* on this branch, a manually-set `param.waypoint.intervals` of 86, without override_intervals on, was output unchanged, but the `snapshot.waypoint.intervals` was 27, as was the correct heuristic value. 

This demonstrates that the parameter section is no longer getting modified, but the snapshot section does still communicate the correct interval counts.

docs/contributing/generation-invariants.md
calcmogul, line 1: The implementation should link to this in relevant places so it's more likely to get updated.

PR #1262, opened 2025-05-30
[trajoptlib] Fix parameter name in Pose2.rotate_by

spacey-sooty:


calcmogul (commented):
What about the other instances of `other`?

spacey-sooty:
> What about the other instances of `other`?

I think its fine for operator overloading, I'll replace it on other geometry methods though.

spacey-sooty:
Most of the rest of the instances of other I've found its been the same type in the parameter. the reason I changed these two is because other implies to me its the same type.

PR #1269, opened 2025-06-23
Port ChoreoLib 2025.0.3 to SystemCore/WPILib 2027

shueja:
Somehow this needs to get published to the Artifactory once it's approved, but it shouldn't be committed to main.


src-tauri/gen/schemas/windows-schema.json
calcmogul, line 1: Stuff in src-tauri/gen is already excluded by .gitignore. I assume this made it in before you merged main.

calcmogul (approved):

PR #1294, opened 2025-08-07
Add support for rotated rectangle constraints

### LINKED ISSUE #1293: Support rotated keep-in rectangles
The keep-in rectangle constraint is represented by a keep-in polygon, so we can do whatever transformation we want before then.

https://github.com/SleipnirGroup/Choreo/blob/main/src-core/src/generation/transformers/constraints.rs#L134-L141
```rs
                ConstraintData::KeepInRectangle { x, y, w, h } => {
                    let xs = vec![x, x + w, x + w, x];
                    let ys = vec![y, y, y + h, y + h];
                    match to_opt {
                        None => generator.wpt_keep_in_polygon(from, xs, ys),
                        Some(to) => generator.sgmt_keep_in_polygon(from, to, xs, ys),
                    }
                },
```

TheTripleV:


calcmogul (changes requested):
The input boxes got cut off because "Rotation" made the boxes render further to the right:
Screenshot_20250807_153530

calcmogul (changes requested):
The rotate handle is a bit awkward because it overlaps with the other corners. Also, dragging a green corner of that rotated rectangle up results in the corner of the box deviating from where the mouse pointer is, like there's some kind of scaling issue.

calcmogul (changes requested):
If I drag a corner into the center of the box (so the box is effectively zero size), it starts dragging the whole box along. When I expand the box again, the mouse is no longer on top of the selected corner. When I do this on main, it just lets the corner pass through the center and let the area expand the other way.

The field-wide keep-in rectangle needs its placement corrected.

calcmogul:
A minimum size equal to the robot's dimensions would make sense.

calcmogul:
The min size enforcement works, but the box corners still move around when they shouldn't at minimum size, and the mouse pointer location still deviates from the drag point.

calcmogul (changes requested):

calcmogul (changes requested):
The corner resize seems to get stuck in some situations.

https://github.com/user-attachments/assets/6ccf549d-c0ea-4d33-addd-ece94a997dc3

PR #1349, opened 2025-11-15
Add AdvantageKit auto chooser

Daniel1464:
This was always a thing that we should have added to the docs, so here it is.
Also, I fixed a couple of small things related to the AutoChooser too.

### DIFF
--- a/docs/choreolib/auto-factory.md
+++ b/docs/choreolib/auto-factory.md
@@ -317,51 +317,69 @@ Choreo's base `AutoChooser` does not support log replay. AdvantageKit users shou
     `AutoChooserAK` will automatically register the auto chooser to smart dashboard, meaning that
     a manual `SmartDashboard.putData` call is no longer necessary.
 
-```java title="AutoChooserAK.java"
-package your.utilities.folder;
+```java title="LoggedAutoChooser.java"
+package your.lib.folder;
 
 import choreo.auto.AutoRoutine;
+import edu.wpi.first.wpilibj.DriverStation;
 import edu.wpi.first.wpilibj2.command.Command;
 import edu.wpi.first.wpilibj2.command.Commands;
+import edu.wpi.first.wpilibj2.command.button.Trigger;
 import org.littletonrobotics.junction.networktables.LoggedDashboardChooser;
 
 import java.util.HashMap;
+import java.util.Optional;
 import java.util.function.Supplier;
 
-/** A replacement for {@link choreo.auto.AutoChooser} that works with AdvantageKit. */
-public class AutoChooserAK {
+/** A replacement for choreo's {@link choreo.auto.AutoChooser} that works with AdvantageKit. */
+public class LoggedAutoChooser {
+    private static final String NONE_NAME = "No Auto";
+
     private final LoggedDashboardChooser<String> impl;
     private final HashMap<String, Supplier<Command>> choices = new HashMap<>();
+    private Optional<DriverStation.Alliance> alliance = Optional.empty();
     private Command selectedAuto = Commands.none();
 
-    /** Creates an {@link AutoChooserAK}. */
-    public AutoChooserAK(String name) {
+    public LoggedAutoChooser(String name) {
         impl = new LoggedDashboardChooser<>(name);
-        impl.onChange(option -> selectedAuto = choices.get(option).get());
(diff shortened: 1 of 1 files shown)


choreolib/src/main/java/choreo/auto/AutoChooser.java
```diff
@@ -45,7 +46,7 @@ public class AutoChooser implements Sendable {
   private final HashMap<String, Supplier<Command>> autoRoutines =
       new HashMap<>(Map.of(NONE_NAME, Commands::none));
 
-  private String selected = NONE_NAME;
+  protected String selected = NONE_NAME;
```
calcmogul: Why was this changed to protected? If you want it published somewhere, then make AutoChoooser work with SmartDashboard directly, like SendableChooser does.

Also, SendableChooser provides a public getter `getSelected()`. Protected members are a code smell.

choreolib/src/main/java/choreo/auto/AutoChooser.java
```diff
@@ -187,7 +188,7 @@ public Command selectedCommandScheduler() {
    * @return The currently selected command.
    */
   public Command selectedCommand() {
-    if (RobotBase.isSimulation() && nameAtGeneration == NONE_NAME) {
+    if (RobotBase.isSimulation() && Objects.equals(nameAtGeneration, NONE_NAME)) {
```
calcmogul:
```suggestion
    if (RobotBase.isSimulation() && nameAtGeneration.equals(NONE_NAME)) {
```

docs/choreolib/auto-factory.md
```diff
@@ -309,3 +309,43 @@ public class Robot extends TimedRobot {
...
+
+```java title="AutoChooserAK.java"
+package your.utilities.folder;
+
+import edu.wpi.first.wpilibj.smartdashboard.SmartDashboard;
+import org.littletonrobotics.junction.LogTable;
+import org.littletonrobotics.junction.Logger;
+import org.littletonrobotics.junction.inputs.LoggableInputs;
+
+public class AutoChooserAK extends AutoChooserImpl {
```
shueja: Is AutoChooserImpl the right thing to extend here? Also this example needs to import AutoChooser, no?

docs/choreolib/auto-factory.md
```diff
@@ -309,3 +309,72 @@ public class Robot extends TimedRobot {
...
+/** A replacement for {@link choreo.auto.AutoChooser} that works with AdvantageKit. */
+public class AutoChooserAK {
+    private final LoggedDashboardChooser<String> impl;
+    private final HashMap<String, Supplier<Command>> choices = new HashMap<>();
+    private Command selectedAuto = Commands.none();
+
+    /** Creates an {@link AutoChooserAK}. */
+    public AutoChooserAK(String name) {
+        impl = new LoggedDashboardChooser<>(name);
+        impl.onChange(option -> selectedAuto = choices.get(option).get());
```
shueja: This doesn't preserve one of the main intents of the AutoChooser, namely that it ensures auto routine generators are not called before the DS is connected and the alliance is available, and that they are regenerated when the alliance changes.

shueja (changes requested):

DJ-Laser:
My take on a LoggedAutoChooser: https://gist.github.com/DJ-Laser/e0c11d7b835489ff63e6e4d2f72e089f
Since the default autoChooser has a special behavior on sim, it's important to keep the command name that will actually be ran logged instead of the user selected name, otherwise replay and real behavior could differ if an auto is selected and ran without alliance info.

Daniel1464:
> My take on a LoggedAutoChooser: https://gist.github.com/DJ-Laser/e0c11d7b835489ff63e6e4d2f72e089f Since the default autoChooser has a special behavior on sim, it's important to keep the command name that will actually be ran logged instead of the user selected name, otherwise replay and real behavior could differ if an auto is selected and ran without alliance info.

Your implementation does seem correct, but isn't alliance info automatically replayed as well? If no alliance info is given, the auto should default to a null-op.
My main concern with your implementation, though, is that its a little bit of a leaky abstraction. Manual select() calls on the AutoChooser will cause  replay to be inaccurate, because they override the AutoChooser's internal selected String property.

PR #1386, opened 2025-12-26
Java identifier UI validation

shueja:
image


shueja:
TODO:
[ ] Decide which math.js reserved words (constants, units, functions) we want to prevent being clobbered, vs which ones we want to allow users to overwrite with variables.
[ ] Expand the name error feedback in genTrajDataFile.ts to genVarDataFile.ts

Daniel1464:
I tried it out on my computer, and right now the variable name seems to wipe automatically, even when it's a valid identifier.

PR #1393, opened 2025-12-29
Store robot config snapshot in .traj; compare for stale trajectory check

shueja:
This is a schema upgrade to trajectory version 3

It adds the `trajectory.config` field, which is a `double` version of RobotConfig:
```
"config":{
    "frontLeft":{
        "x":0.2794,
        "y":0.2794
    },
    "backLeft":{
        "x":-0.2794,
        "y":0.2794
    },
    "mass":68.0388555,
    "inertia":6.0,
    "gearing":6.5,
    "radius":0.0508,
    "vmax":628.3185307179587,
    "tmax":1.2,
    "cof":1.5,
    "bumper":{
        "front":0.4064,
        "side":0.4064,
        "back":0.4064
    },
    "differentialTrackWidth":0.5588
  },
```

It compares this snapshot against the current robot config when either the .traj or the current robot config snapshot change. If the two copies are different, the trajectory should be out of date. Expressions in the robot config that calculate to the same float value are considered equivalent for this check, since it's just comparing the float values.
NOTE: changes that don't necessarily affect the trajectory are still considered cause for regeneration. This includes differentialTrackWidth on swerve trajectories, motor torque limits which are superseded by CoF limits, etc.

### DIFF
--- a/src/document/tauriCommands.ts
+++ b/src/document/tauriCommands.ts
@@ -8,6 +8,11 @@ export const Commands = {
     invoke<number[]>("guess_control_interval_counts", { config, trajectory }),
 
   /**
+   * @returns The estimate of available parallelism from Rust's std::thread
+   */
+  getWorkerCount: () => invoke<number>("get_worker_count"),
+
+  /**
    * Generates a `Trajectory` using the specified `Project` and `Trajectory`.
    *
    * @param project The `Project` to use for generation.
--- a/choreolib/py/choreo/trajectory/__init__.py
+++ b/choreolib/py/choreo/trajectory/__init__.py
@@ -1,20 +1,148 @@
 from __future__ import annotations
 
-import json
 import math
-import os
 from dataclasses import dataclass
 from typing import TypeGuard
 
 import numpy as np
 from choreo.util import DEFAULT_YEAR, get_flipper_for_year
-from scipy.integrate import solve_ivp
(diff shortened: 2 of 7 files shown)


src/document/tauriCommands.ts
```diff
@@ -130,6 +130,14 @@ export const Commands = {
...
     invoke<boolean>("trajectory_up_to_date", { trajectory }),
+
+  /**
+   * Returns if the `Trajectory` robot config snapshot is are equivalent.
+   * @param trajectory The `Trajectory` to check
+   * @returns true if the parameters and snapshots are equivalent, false if not.
+   */
+  configMatches: (config1: RobotConfig<number>, config2: RobotConfig<number>) =>
+    invoke<boolean>("config_matches", { config1, config2 }),
   /**
```
calcmogul:
```suggestion
   * Returns if the `Trajectory` robot config snapshot is are equivalent.
   * 
   * @param trajectory The `Trajectory` to check
   * @returns true if the parameters and snapshots are equivalent, false if not.
   */
  configMatches: (config1: RobotConfig<number>, config2: RobotConfig<number>) =>
    invoke<boolean>("config_matches", { config1, config2 }),

/**
```

calcmogul (changes requested):

calcmogul (approved):
Modifying the document settings causes the "generate all outdated" button to do the correct thing.

PR #1394, opened 2025-12-30
Fix bug with unintentionally mutating parameters in CLI

shueja:
Refactor data flow to calculate the interval counts before generation and send it as a progress update instead of a preflight command invocation.

(The preflight invocation was broken anyway because it didn't do all the parameter adjustment that affects control interval counts estimation)

Rework pre-generation parameter adjustment to work on a snapshot of the parameters which is passed to generation logic but does not mutate the original TrajectoryFile struct.

Fix remote generation completion so that the remote process puts together the full final TrajectoryFile instead of just sending back the samples.

Move all preprocess and postprocess steps into either the TrajectoryFileGenerator::new or ::generate to consolidate generation steps.

NOTE: this assumes the path in the UI is not edited while generation is running, which is a condition we plan to enforce but is not yet mechanically enforced


shueja:
TODO separately: whatever parts of #1257 aren't superseded by this (I.e the GenerationInvariant trait is still a good idea)

shueja:
Yes

PR #1396, opened 2026-01-01
Code-deduplicate the conversion from path uuid to handle

shueja:


Daniel1464:
I think it looks good

PR #1401, opened 2026-01-04
Limit number of parallel generation jobs

### LINKED ISSUE #561: Add button that generates all trajectories at once in parallel

shueja:
Closes #561

This is not an overall solution, it just changes the implementation of the GUI generate-all buttons to spawn a limited set of workers that take the next generation task when they become available.

### DIFF
--- a/src-tauri/src/api.rs
+++ b/src-tauri/src/api.rs
@@ -272,6 +272,7 @@ pub fn open_diagnostic_file(
         Err(ChoreoError::FileNotFound(None).into())
     }
 }
+
 #[tauri::command]
 pub fn get_worker_count() -> TauriResult<NonZero<usize>> {
     // if this unwrap panics, 4 is equal to 0 and we have bigger problems.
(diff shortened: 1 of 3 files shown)


src/document/DocumentModel.tsx
```diff
@@ -206,12 +227,13 @@ export const DocumentStore = types
       pathStore.markers.forEach((m) => {
         m.from.setTrajectoryTargetIndex(m.from.getTargetIndex());
       });
-      pathStore.ui.setGenerating(true);
+      //pathStore.ui.setGenerating(true);
```
calcmogul: Remove this commented out line.

src/document/tauriCommands.ts
```diff
@@ -6,7 +6,7 @@ export type ChoreoResult<T> = T | ChoreoError;
 export const Commands = {
   guessIntervals: (config: RobotConfig<Expr>, trajectory: Trajectory) =>
     invoke<number[]>("guess_control_interval_counts", { config, trajectory }),
-
+  getWorkerCount: () => invoke<number>("get_worker_count"),
```
calcmogul: Looks like the intent was to put an empty line between each declaration, so there should be empty lines before and after this one.

src-tauri/src/api.rs
```diff
@@ -272,3 +272,8 @@ pub fn open_diagnostic_file(
         Err(ChoreoError::FileNotFound(None).into())
     }
 }
+#[tauri::command]
```
calcmogul: Put an empty newline before this.

PR #1406, opened 2026-01-06
Fix renaming a variable causing modification of every expression in every file

shueja:
By reassigning to every expression even if its contents didn't change, we triggered the UI to consider it modified.


shueja (commented):
Tested that with this change, only paths using the given variable are re-saved, with the updated variable name as expected.

PR #1409, opened 2026-01-08
Fix error feedback for schema being too new

shueja:
This will fail on the first too-new file with a notice to update Choreo


shueja:
image

PR #1412, opened 2026-01-09
Make addCmd() and addRoutine() return this for chaining purposes

### LINKED ISSUE #1411: AutoChooser.addCmd/addRoutine should return "this" for chaining.

Daniel1464:
Resolves #1411.


calcmogul (changes requested):
Just needs some `@return` doc tags.

PR #1436, opened 2026-02-03
Move comment in genTrajDataFile.ts to avoid breaking formatters

### LINKED ISSUE #1435: Comment in generated ChoreoTraj file breaks formatters
In the generated ChoreoTraj.java there is a comment in the middle of the import section.
The Spotless Java formatter fails when it encounters this.

jthro:
Resolves #1435


calcmogul:
Could you provide a minimal reproducible example? When I add the existing comment to a file in wpimath, allwpilib's spotless formatter config handles it fine (i.e., I'm unable to reproduce your issue).

PR #1475, opened 2026-04-21
Implement better motor model

### LINKED ISSUE #1245: Better motor model
Ideally we could model the motors better within trajoptlib to prevent the user from having to find the values needed (either through a motor curve or trial and error). Ideally the end user API would take certain motor specs and things like current limits and we could model from there. This could also allow us to bring in presets for common motors where the user only has to configure things like current limits.

spacey-sooty:
Resolves #1245


trajoptlib/src/swerve_trajectory_generator.cpp
```diff
@@ -241,30 +245,47 @@ SwerveTrajectoryGenerator::SwerveTrajectoryGenerator(
...
+      Translation2v<double> v_volts_wrt_robot = v_wheel_wrt_robot * kV;
+      auto proj_v_F = v_wheel_wrt_robot * (F_wrt_robot.dot(v_wheel_wrt_robot)) /
+                      (v_wheel_wrt_robot.squared_norm());
+
+      const auto omega =
+          v_wheel_wrt_robot.norm() / path.drivetrain.wheel_radius;
+      auto I_stator = problem.decision_variable();
+      auto v = I_stator * R + omega * kV;
+
+      problem.subject_to(slp::abs(v) <= v_supply);
```
calcmogul: Instead of using one abs() constraint, which is nonlinear, use a bound constraint, which creates two linear constraints.
```suggestion
      problem.subject_to(slp::bounds(-v_supply, v, v_supply));
```

trajoptlib/src/swerve_trajectory_generator.cpp
```diff
@@ -241,30 +245,47 @@ SwerveTrajectoryGenerator::SwerveTrajectoryGenerator(
...
-
-      // |F|₂² ≤ Fₘₐₓ²
-      problem.subject_to(module_force.squared_norm() <= F_max * F_max);
+      auto F_wrt_robot = module_force.rotate_by(-θ_k);
+
+      const auto kV = path.drivetrain.motor_config.kV;
+      const auto R = path.drivetrain.motor_config.resistance();
+      const auto I_supply_limit = path.drivetrain.motor_config.supply_limit;
+      const auto I_stator_limit = path.drivetrain.motor_config.stator_limit;
+      const auto v_supply = 12;
```
calcmogul: Voltage should use `V`. `v` already means velocity.
```suggestion
      const auto V_supply = 12;
```

trajoptlib/src/swerve_trajectory_generator.cpp
```diff
@@ -241,30 +245,47 @@ SwerveTrajectoryGenerator::SwerveTrajectoryGenerator(
...
+      auto proj_v_F = v_wheel_wrt_robot * (F_wrt_robot.dot(v_wheel_wrt_robot)) /
+                      (v_wheel_wrt_robot.squared_norm());
+
+      const auto omega =
+          v_wheel_wrt_robot.norm() / path.drivetrain.wheel_radius;
+      auto I_stator = problem.decision_variable();
+      auto v = I_stator * R + omega * kV;
+
+      problem.subject_to(slp::abs(v) <= v_supply);
+      problem.subject_to(slp::abs((2 * v - omega * kV) * (2 * v - omega * kV) -
```
calcmogul: This will need a derivation or paper link eventually.

calcmogul (changes requested):

PR #1485, opened 2026-05-17
C++-based Choreo generator program

shueja:
The goal here is to avoid Rust-C++ FFI in the portion of Choreo that translates .chor and .traj into TrajoptLib calls. This depends on wpiutil and wpimath for maintainability and reuse of existing resources. The `choreo_core/include/data` directory should be compilable with WITH_TRAJOPT not defined, in which case it will not depend on TrajoptLib and can be used on-robot.

Not marking it as a draft, even though it isn't ready, because I want review comments on the code architecture.


choreo_core/include/data/expr.hpp
```diff
@@ -0,0 +1,60 @@
...
+  /// valid state for an Expr. Instead, use the constructor that accepts a
+  /// BaseUnit or a double, which will properly initialize the exp string.
+  Expr() = default;
+  Expr(std::string exp, double val) : exp(exp), val(val) {}
+  /// a constructor that accepts anything convertible to BaseUnit directly and
+  /// formats the string with the unit abbreviation
+  template <typename U,
+            std::enable_if_t<std::is_convertible_v<U, BaseUnit>, int> = 1>
+  // NOLINTNEXTLINE (google-explicit-constructor)
+  Expr(U&& u) : Expr(static_cast<BaseUnit>(std::forward<U>(u))) {}
```
calcmogul:
```suggestion
  template <typename U>
    requires std::is_convertible_v<U, BaseUnit>
  // NOLINTNEXTLINE (google-explicit-constructor)
  Expr(U&& u) : Expr(static_cast<BaseUnit>(std::forward<U>(u))) {}
```

document/cpp/include/choreo/variables/dimension.hpp
```diff
@@ -0,0 +1,137 @@
...
+#include <wpi/units/torque.hpp>
+#include <wpi/units/velocity.hpp>
+#include <wpi/units/voltage.hpp>
+
+namespace choreo {
+namespace dimensions {
+
+template <typename T>
+struct Dimension {
+  inline static const std::string tag = "default value";
```
calcmogul: You can probably use `inline static constexpr std::string_view` here since the string literals are constexpr and have global storage duration.

shueja:
This PR makes schema changes but I do not intend to offer an upgrade path because the field coordinate system is changing. 

Notable schema changes:

* Silent breakage: coordinate system
  * Following WPILib, the field coordinate system is now (0,0) at field center with +x going towards the BLUE drivers instead of the RED drivers.

* Samples/Sample Lists:
  * To leverage https://github.com/wpilibsuite/allwpilib/pull/8172, sample structure is fully changing. Not necessarily seen in code yet.

* Robot config
  * now takes a list of Translations for bumper corners and wheel locations
  * motor model parameters reorganized in anticipation of current-based motor modeling
* New variable types added
  * Translation (x: Length, y:Length). This is the same structure used for bumper corners and wheel locations in the robot config.
  * Region (x: Length, y:Length, heading: Angle, w: Length, h: Length). Depending on where this is used, it can represent an ellipse or a rectangle, with axis-rotation supported.
* New expression dimensions added
  * Current (Amps)
  * KT (N*m per Amp)
  * KV (Volt*seconds per radian)
  * TBD: should we add Volts?
* Constraints
  * KeepInLane now includes coordinates for its center line, with an option to have the solver use the constraint range start and end points. This increases flexibility while avoiding constraints requiring the context of the range start and end positions.
  * Planned to add Keep In Rect and Ellipse with axis-rotation support. KeepInCircle likely remains a separate constraint because of its special UI behavior and TrajoptLib setup.
* Snapshot
  * Now is a full copy of parameters and robot config, not a version that converts Exprs to doubles. This takes more serialization space but allows reversion to the last generated settings.
* Event marker
  * PathPlannerLib command association removed for now, @mjansen4857 has not yet decided what that will look like for 2027

PR #1501, opened 2026-07-09
Upgrade to tsc 6.0.3

calcmogul:


src/document/UIStateStore.tsx
```diff
@@ -16,10 +16,10 @@ import {
   navbarIndexToConstraint,
   navbarIndexToConstraintDefinition,
   navbarIndexToConstraintKey
-} from "./UIData";
-import { tracing } from "./tauriTracing";
-import { CustomFieldData } from "./schema/DocumentTypes";
-import { defaultFieldData } from "../components/field/svg/fields/JSONFieldImage";
+} from "./UIData.ts";
+import { tracing } from "./tauriTracing.ts";
```
Daniel1464: Does TS 7 have a new requirement on adding .ts extensions to these imports, or is it a separate config change?

calcmogul: I was getting this error:
```
src/Body.tsx:3:20 - error TS2835: Relative import paths need explicit file extensions in ECMAScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Did you mean './components/navbar/Navbar.js'?

3 import Navbar from "./components/navbar/Navbar";
                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

tsconfig.json
```diff
@@ -1,20 +1,20 @@
...
-    "strict": true,
     "forceConsistentCasingInFileNames": true,
-    "module": "ESNext",
-    "moduleResolution": "Node",
-    "resolveJsonModule": true,
     "isolatedModules": true,
+    "jsx": "react-jsx",
+    "lib": ["DOM", "DOM.Iterable", "ESNext"],
+    "module": "nodenext",
+    "moduleResolution": "nodenext",
```
Gold856: You probably don't want this, since your TypeScript isn't running on Node.js. [Use `esnext` for `module` and `bundler` for `moduleResolution`](https://www.typescriptlang.org/docs/handbook/modules/reference.html#es2015-es2020-es2022-esnext). This also means [you don't need file extensions on relative imports](https://www.typescriptlang.org/tsconfig/#moduleResolution).

Gold856 (changes requested):

